PR #1, opened 2016-03-22
Replace Future#promise with Future#then

Avaq:
So I've been having fun with Fluture. At one point I implement `Future#promise` to allow casting to Promise for APIs that consume them, and basically an alternative to `fork`. Today I realized I may as well implement `then`, using the `Future.cache` I already had, combined with the `Future.promise` I already had. Now I can pass Future to APIs that consume promises. And it turns out it passes the Promise/A+ test suite (granted I'm still returning a Promise and not a Future). The downside is that I've introduced a line of code which mutates shared state. The advantage is that `then` is a standardized interface for pulling the value out, as opposed to `fork`.

### DIFF
--- a/fluture.js
+++ b/fluture.js
@@ -132,8 +132,8 @@
     if(!isFunction(f)) error$invalidArgument('Future#value', 0, 'be a function', f);
   }
 
-  function check$promise(it){
-    if(!isFluture(it)) error$invalidContext('Future#promise', it);
+  function check$then(it){
+    if(!isFluture(it)) error$invalidContext('Future#then', it);
   }
 
   function check$cache(m){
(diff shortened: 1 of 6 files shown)


fluture.js
```diff
@@ -309,12 +309,12 @@
...
 
-  function Future$promise(){
-    check$promise(this);
-    const _this = this;
-    return new Promise(function Future$promise$do(resolve, reject){
-      _this.fork(reject, resolve);
-    });
+  function Future$then(onResolve, onReject){
+    check$then(this);
+    const cached = this._cached || (this._cached = Future.cache(this));
```
Avaq: This mutation is a shame.

PR #10, opened 2016-04-26
Fix a common bug with do-notation

Avaq:
Commit message:

> Because of the way iterators are "depleted" upon use,
the returned Future could only safely be forked once.

> This is something which is known within the community
to be unfixable in a general way. However, because
Future.do is specialized, I was able to fix it by
respawning the iterator upon every new fork().

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,8 +98,8 @@ all types used within these signatures follows:
 
 The Future constructor. Creates a new instance of Future by taking a single
 parameter `fork`: A function which takes two callbacks. Both are continuations
-for an asynchronous computation. The first is `reject`, commonly abreviated to
-`rej`. The second `resolve`, which abreviates to `res`. The `fork` function is
+for an asynchronous computation. The first is `reject`, commonly abbreviated to
+`rej`. The second `resolve`, which abbreviates to `res`. The `fork` function is
 expected to call `rej` once an error occurs, or `res` with the result of the
 asynchronous computation.
 
@@ -362,7 +362,7 @@ Future.of(x => x + 1)
 
 #### `race :: Future a b ~> Future a b -> Future a b`
 
-Race two Futures against eachother. Creates a new Future which resolves or
+Race two Futures against each other. Creates a new Future which resolves or
 rejects with the resolution or rejection value of the first Future to settle.
 
 ```js
@@ -376,7 +376,7 @@ Future.after(100, 'hello')
 
 Logical or for Futures.
 
-Returns a new Future which either resolves with the first resolutation value, or
+Returns a new Future which either resolves with the first resolution value, or
 rejects with the last rejection value once and if both Futures reject.
 
 This behaves analogues to how JavaScript's or operator does, except both
@@ -578,7 +578,10 @@ Returns true for [Forkables](#type-signatures) and false for everything else.
 #### `do :: (() -> Iterator) -> Future a b`
 
 A specialized version of [fantasy-do][19] which works only for Futures, but has
-the advantage of type-checking and not having to pass `Future.of`.
+the advantage of type-checking and not having to pass `Future.of`. Another
+advantage is that the returned Future can be forked multiple times, as opposed
+to with a general `fantasy-do` solution, where forking the Future a second time
+behaves erroneously.
 
 Takes a function which returns an [Iterator](#type-signatures), commonly a
 generator-function, and chains every produced Future over the previous.
--- a/fluture.js
+++ b/fluture.js
@@ -717,14 +717,16 @@
 
   Future.do = function Future$do(f){
     check$do(f);
-    const g = f();
-    check$do$g(g);
-    const next = function Future$do$next(x){
-      const o = g.next(x);
-      check$do$next(o);
-      return o.done ? Future$of(o.value) : o.value.chain(next);
-    };
-    return next();
+    return new FutureClass(function Future$do$fork(rej, res){
+      const g = f();
+      check$do$g(g);
+      const next = function Future$do$next(x){
+        const o = g.next(x);
+        check$do$next(o);
+        return o.done ? Future$of(o.value) : o.value.chain(Future$do$next);
+      };
+      return next().fork(rej, res);
+    });
   };
 
   return Future;
--- a/test/fluture.test.js
+++ b/test/fluture.test.js
@@ -1500,19 +1500,21 @@ describe('Other', () => {
 
     it('throws TypeError when the given function does not return an interator', () => {
       const xs = [NaN, {}, [], 1, 'a', new Date, undefined, null, () => {}, {next: 'hello'}];
-      const fs = xs.map(x => () => Future.do(() => x));
+      const fs = xs.map(x => () => Future.do(() => x).fork(noop, noop));
       fs.forEach(f => expect(f).to.throw(TypeError, /Future/));
     });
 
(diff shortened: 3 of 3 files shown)


Avaq:
> This is something which is known within the community to be unfixable in a general way

Reference: https://github.com/DrBoolean/freeky/issues/4#issue-149842139

---

However, I did just come up with a way to solve it generically, I think: http://goo.gl/VlMI2g

davidchambers:
I see that we're now wrapping with this:

```javascript
new FutureClass(function Future$do$fork(rej, res){…});
```

We also replaced `next()` with `next().fork(rej, res)`.

Could you explain how changing the context of `f()` in this way solves the problem? I can see how it might, but my thoughts are fuzzy.

Avaq:
The problem is that the iterator (`g`) is stateful and shared between every `fork` of the Future. So when forking the Future the first time, the iterator will start depleting (by having its `next` method called). If you then `fork` *the same* Future again, the shared iterator will already have been (partially) depleted.

The solution in this case is to force a new iterator to be spawned every time the resulting Future is forked. We can achieve this by wrapping the whole thing, *including the creation of the iterator* inside the `fork` function of our final Future.

Calling `next().fork(rej, res)` just makes it so the state of the inner Future is transferred to the outer Future.

davidchambers:
That makes sense. How do we access iterator values beyond the first? It seems we define `g` as the result of `f()`, but invoke `g.next` exactly once.

Avaq:
`g.next` is invoked every time `Future$do$next` (alias `next`) is called. `next` is called recursively by passing it to `o.value.chain`.

davidchambers:
Ah, yes. I see it now. Thanks for the explanation. :)

fluture.js
```diff
@@ -717,14 +717,16 @@
...
-      return o.done ? Future$of(o.value) : o.value.chain(next);
-    };
-    return next();
+    return new FutureClass(function Future$do$fork(rej, res){
+      const g = f();
+      check$do$g(g);
+      const next = function Future$do$next(x){
+        const o = g.next(x);
+        check$do$next(o);
+        return o.done ? Future$of(o.value) : o.value.chain(next);
```
davidchambers: I'd prefer to see `o.value.chain(Future$do$next)` in place of `o.value.chain(next)`. Referencing a function by name from within its body is more self-contained than referencing a variable in outer scope.

README.md
```diff
@@ -578,7 +578,10 @@ Returns true for [Forkables](#type-signatures) and false for everything else.
 #### `do :: (() -> Iterator) -> Future a b`
 
 A specialized version of [fantasy-do][19] which works only for Futures, but has
-the advantage of type-checking and not having to pass `Future.of`.
+the advantage of type-checking and not having to pass `Future.of`. Another
+advantge is that the returned Future can be forked multiple times, as opposed
```
davidchambers: `s/advantge/advantage/`

README.md
```diff
@@ -578,7 +578,10 @@ Returns true for [Forkables](#type-signatures) and false for everything else.
...
+to with a general `fantasy-do` solution, where forking the Future a second time
+would introduce erronous behaviour.
```
davidchambers: `s/erronous/erroneous/`

kedashoe: erronous -> erroneous? Maybe "unpredictable" is better?

kedashoe: ah! you got me that time @davidchambers  ;)

fluture.js
Avaq: Good point! The same thought crossed my mind whilst typing "alias `next`" to you earlier.

README.md
Avaq: It's not only unpredictable, it's actually just wrong. Sometimes it even leads to reference errors and stuff like that.

Avaq:
Thanks guys! I've corrected the grammar and typo's you pointed out (and some others while I was at it).

fluture.js
```diff
@@ -717,14 +717,15 @@
...
-    };
-    return next();
+    return new FutureClass(function Future$do$fork(rej, res){
+      const g = f();
+      check$do$g(g);
+      return (function Future$do$next(x){
+        const o = g.next(x);
+        check$do$next(o);
+        return o.done ? Future$of(o.value) : o.value.chain(Future$do$next);
+      }()).fork(rej, res);
```
Avaq: What do you think of in-lining it like this @davidchambers?

davidchambers: I don't mind either way. I think it's a little easier to read without the IIFE, though.

davidchambers:
:deciduous_tree:

fluture.js
Avaq: The process leading upto this was as follows:

1. I removed the `const next = ` part
2. `return next()` would have to be changed
3. but `function Future$do$fork(){ … }` is a function declaration now, which make me feel *iffy*
4. I realized an IIFE would cause it to be a function expression again, coincidentally; 
scrot_20160427-114608_310x275

I'm not sure now..

davidchambers: Haha! I suggest reverting the change, but using `Future$do$next` rather than `next` inside the function.

Avaq: Marvelous suggestion!

safareli:
that's nice! 👍
Will try to adopt this in [DrBoolean/freeky](https://github.com/DrBoolean/freeky)

One thing to note is that if user `yield`-s `List` monad (or any monad which calls function passed to `chain` multiple times), will not work properly.

Avaq:
> that's nice!

Hi @safareli, I'm assuming you were talking about [the generic solution](http://goo.gl/VlMI2g)?

> any monad which calls function passed to chain multiple times [...] will not work properly 

Of course! I had a feeling it couldn't be as simple as wrapping it in `chain`, I just couldn't figure out why. This clears that up! :D

safareli:
Yes I was talking about that generic solution. I twiked that to work with [Free](https://github.com/DrBoolean/freeky/pull/7/files#diff-cd5e599fb139f969c5060df4d0905c7aR4)
 
The problem is that calling `next` changes state of iterator so calling `next` multiple times changes it's state multiple times. For example in here [Do.Multi](https://github.com/russellmcc/fantasydo/blob/master/index.js#L18) generator is replayed multiple times for it to works with arays, but it's specific to arrays, and it will not work properly with monads which performe side effects when`yield`-ed like IO.

As Future is only calling `f` passed to chain only once it will not have this kind of issues i think. but it is worth to note in documentation to not `yield` such monads which call `f` passed to chain multiple times.

> Wouldn't the same thing break with all existing `Do` implementations?

Yes

Avaq:
> Yes

So I guess my `of(nil).chain(_ => ...)` solution is a general improvement over what the JS community has been using to this point. Maybe we should inform the maintainers of [`fantasydo`](https://github.com/russellmcc/fantasydo), [`fantasy-do`](https://github.com/jwoudenberg/fantasy-do) and [`do-notation`](https://github.com/Risto-Stevcev/do-notation).

safareli:
yap, your `of(_).chain(_ => ...)` is definitely good thing as you could reuse result of `do`. 


from [russellmcc/fantasydo](https://github.com/russellmcc/fantasydo):
> If your monad only calls the argument to the chain function once ('non-branching'), you can use Fantasy Do's normal, high-performance mode.

It looks like `fantasydo` does understand that, but it does not work as you have proposed, so in order for it's `do` to return something it runs generator to get first `yield`-ed object and returns it, so it's result is not reusable.

PR #14, opened 2016-05-17
Implement finally()

Avaq:
This commit will add a new function:

```js
.finally :: Future a b -> Future a c -> Future a c
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -350,6 +350,36 @@ Future.reject('it broke')
 //> Left('it broke')
 ```
 
+#### finally
+##### `#finally :: Future a b ~> Future a c -> Future a b`
+##### `.finally :: Future a c -> Future a b -> Future a b`
+
+Run a second Future after the first settles (successfully or unsuccessfully).
+Rejects with the rejection reason from the first or second Future, or resolves
+with the resolution value from the first Future.
+
+```js
+Future.of('Hello')
+.finally(Future.of('All done!').map(console.log))
+.fork(console.error, console.log)
+//> "All done!"
+//> "Hello"
+```
+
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -350,6 +350,36 @@ Future.reject('it broke')
 //> Left('it broke')
 ```
 
+#### finally
+##### `#finally :: Future a b ~> Future a c -> Future a b`
+##### `.finally :: Future a b -> Future a c -> Future a c`
```
davidchambers: Are these return types different intentionally?

Avaq: The functional API takes the "data" last, whereas in the method API `this` *is* the data.

So:

```js
computation.finally(finallyComputation)
==
Future.finally(finallyComputation, computation)
```

The different output type is meant to draw attention to that fact.

Avaq: Perhaps it'd be better to leave the output type the same `Future a b` and switch `b` and `c` between the first two types.

davidchambers: I like that idea! Having the return types match would make it clear that the arguments are flipped (and that there's no typo).

README.md
```diff
@@ -350,6 +350,36 @@ Future.reject('it broke')
 //> Left('it broke')
 ```
 
+#### finally
+##### `#finally :: Future a b ~> Future a c -> Future a b`
+##### `.finally :: Future a c -> Future a b -> Future a b`
+
+Run a second Future after the first ends, despite of it rejecting or resolving.
```
davidchambers: This doesn't quite make sense. I suggest:

_Run a second Future after the first ends (successfully or unsuccessfully)._

Avaq: Good idea

PR #15, opened 2016-06-21
Add Future.hook()

Avaq:
> Much like `chain`, but takes a "cleanup" computation first, which runs
> *after* the second settles (successfully or unsuccessfully). This allows for
> acquired resources to be disposed, connections to be closed, etc.

### DIFF
--- a/fluture.js
+++ b/fluture.js
@@ -226,6 +226,26 @@
     if(!isFunction(g)) error$invalidArgument('Future#fold', 1, 'be a function', g);
   }
 
+  function check$hook(it, f, g){
+    if(!isFuture(it)) error$invalidContext('Future#hook', it);
+    if(!isFunction(f)) error$invalidArgument('Future#hook', 0, 'be a function', f);
+    if(!isFunction(g)) error$invalidArgument('Future#hook', 1, 'be a function', g);
+  }
+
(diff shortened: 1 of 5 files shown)


fluture.js
```diff
@@ -617,6 +658,19 @@
     return f;
   }
 
+  //Creates a dispatcher for a binary method, but takes the object first rather than last.
+  function createInvertedBinaryDispatcher(method){
+    const f = function invertedBinaryDispatch(m, a, b){
+      if(arguments.length === 1) return unaryPartial(f, m);
+      if(arguments.length === 2) return binaryPartial(f, m, a);
+      if(m && typeof m[method] === 'function') return m[method](a, b);
```
davidchambers: I find all this type checking quite upsetting. :stuck_out_tongue_winking_eye:

Avaq: I'm sorry. ;) It's an attempt to strike a balance between safety and performance.

The `length` checks are the fastest way I could "curry", with the performance loss of just the checks themselves if the function is called with all arguments.

The `typeof` check is just to improve the default error message you would get in that situation: `m.hook is not a function`.

davidchambers: Thanks for the explanation. As you know I'm in favour of descriptive error messages. :)

PR #17, opened 2016-07-10
Add Future.Guarded constructor

Avaq:
There has always been [some confusion](https://github.com/Avaq/Fluture/issues/7) surrounding the fact that the following code: 

```js
Future(rej => setInterval(rej, 100, 'hello')).fork(console.log, console.log)
```

...Will log "hello" to the console continuously. This is especially concerning when wrapping an event-emitter-based API in a Future, because the API might just decide to emit the event multiple times. Now there is `emitter.once()`, but that doesn't help in cases where one event calls the reject-continuation, and another event calls the resolve-continuation:

```js
Future((rej, res) => {
  emitter.once('error', rej).once('done', res);
})
```

When the Future above is forked, it could happen that both continuations are called, and this is in the case where the developer took care to use `once`.

---

One could argue that this behavior is similar to how plain old callbacks work, but there's a reason callbacks are so often associated with hell.

The `Guarded` constructor introduced in this pull-request simply "guards" the continuations for you. Meaning no more continuation will be called after the first has. The goal here is to supply this as an alternative so not to make a breaking change. If the alternative turns out to be practical and has no disadvantages I would like to change the standard `Future` constructor to adopt this new behavior for a `2.0.0` release (which will consequently allow me to remove some safety mechanisms from other parts of the code).

This new behavior does affect performance. The Guarded constructor is twice as slow as the Future constructor (but still 10 times as fast as the Promise constructor). The good news is that internal performance will not be affected. Internal functions don't build on top of the constructor which is exposed to users, but rather a much lighter one which does not have to catch user-land mistakes. So the only time performance will be affected is when the users themselves use the constructor, which I believe is quite rare, and even more rare in hot paths.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,7 @@ getPackageName('package.json')
   1. [Type signatures](#type-signatures)
   1. [Creating Futures](#creating-futures)
     * [Future](#future)
+    * [Guarded](#guarded)
     * [of](#of)
     * [reject](#reject)
     * [after](#after)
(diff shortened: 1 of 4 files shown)


davidchambers:
This seems useful. :)

PR #26, opened 2016-09-15
Support FantasyLand 1.x

Avaq:
* Prefixed method names
* Flipped `ap`

### DIFF
--- a/README.md
+++ b/README.md
@@ -314,18 +314,17 @@ process.on('SIGINT', cancel);
 ```
 
 #### ap
-##### `#ap :: Future a (b -> c) ~> Future a b -> Future a c`
+##### `#ap :: Future a b ~> Future a (b -> c) -> Future a c`
 ##### `.ap :: Apply m => m (a -> b) -> m a -> m b`
 
-Apply the resolution value, which is expected to be a function (as in
-`Future.of(a_function)`), to the resolution value in the given Future. Both
-Futures involved will run in parallel, and if one rejects the resulting Future
-will also be rejected. To learn more about the exact behaviour of `ap`, check
-out its [spec][14].
+Applies the function contained in the right-hand Future to the value contained
+in the left-hand Future. Both Futures involved will run in parallel, and if one
+rejects the resulting Future will also be rejected. To learn more about the
+exact behaviour of `ap`, check out its [spec][14].
 
 ```js
-Future.of(x => x + 1)
-.ap(Future.of(1))
+Future.of(1)
+.ap(Future.of(x => x + 1))
 .fork(console.error, console.log);
 //> 2
 ```
--- a/README.md
+++ b/README.md
@@ -1,7 +1,5 @@
 # Fluture
 
-[<img src="https://raw.github.com/fantasyland/fantasy-land/master/logo.png" align="right" width="196" height="196" alt="Fantasy Land" />][1]
-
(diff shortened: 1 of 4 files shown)


Avaq:
I would like to ask the community whether there's anything I forgot in order to support FL-1.0?

rpominov:
Prefixes and flipped `ap` is probably all there is in Fluture.

fluture.js
```diff
@@ -701,15 +702,6 @@
...
 
-  //Creates a dispatcher for a unary method, but takes the object first rather than last.
-  function createInvertedUnaryDispatcher(method){
-    return function invertedUnaryDispatch(m, a){
-      if(arguments.length === 1) return unaryPartial(invertedUnaryDispatch, m);
-      if(m && typeof m[method] === 'function') return m[method](a);
-      error$invalidArgument(`Future.${method}`, 0, `have a "${method}" method`, m);
-    };
-  }
-
```
davidchambers: It's nice to see this go away. :)

test/fluture.test.js
```diff
@@ -6,11 +6,11 @@ const Readable = require('stream').Readable;
...
-  bimap: 'bimap',
-  chain: 'chain',
-  ap: 'ap',
-  of: 'of'
+  map: 'fantasy-land/map',
+  bimap: 'fantasy-land/bimap',
+  chain: 'fantasy-land/chain',
+  ap: 'fantasy-land/ap',
+  of: 'fantasy-land/of'
 };
```
davidchambers: I suggest adding fantasy-land as a dev dependency and requiring it here to catch typos. :)

Avaq: Ahh, that's a great idea! That way we're ensuring compatibility with the naming conventions of a specific FL version, without having a hard dependency to it! :D

rpominov: Just curious why not actually add a hard dependency?

rpominov: I mean I can understand not wanting a peer dependency, but what bad about a normal dependency?

davidchambers: > Just curious why not actually add a hard dependency?

I don't see compelling reasons to do so: https://github.com/sanctuary-js/sanctuary-type-classes/pull/2#issuecomment-247303007.

Avaq: FantasyLand comes with its own dependency tree which seems to me like unnecessary overhead. It's not much of a big deal, but if there's no practical difference, or even much of a code-reuse advantage, I consider it cleaner not to depend on it.

rpominov: > FantasyLand comes with its own dependency tree which seems to me like unnecessary overhead.

Didn't know that, created a PR that fixes that https://github.com/fantasyland/fantasy-land/pull/168

Other points are valid though, I understand it better now, thanks!

README.md
```diff
@@ -802,3 +807,4 @@ means butterfly in Romanian; A creature you might expect to see in Fantasy Land.
 [22]: https://github.com/Avaq/Fluture/wiki/Comparison-to-Promises
 [23]: https://vimeo.com/106008027
 [24]: https://github.com/fantasyland/fantasy-land#bifunctor
+[25]: https://github.com/fantasyland/fantasy-land/tree/1.0.0
```
davidchambers: Should this be `1.0.1`?

Avaq: I wish it could be `^1.0.0`, maybe I should just use `master` and pin it once a `2.0` comes out.

PR #28, opened 2016-09-17
Add `fs` to readme example

sotojuan:
This might seem kind of silly but it makes the example copy-pastable (copy to editor, save, run), which is very friendly for people trying to library out for the first time.

People will get `fs is missing` errors without it and of course those are easily fixed, but we might as well not have users encounter that.


davidchambers:
Good idea. :)

Avaq:
Thank you for your contribution! It makes perfect sense. :)

PR #31, opened 2016-09-23
Test for Static Land compliance

### LINKED ISSUE #22: Static-Land compliance
As far as I can see, we are already compliant. Just need to write some tests to make sure.

https://github.com/rpominov/static-land

Avaq:
And add a reference to static land in the readme. Closes #22.
Ping @rpominov


rpominov:
Looks great! Don't forget to add it here https://github.com/rpominov/static-land/wiki/Compatible-libraries 😉

PR #33, opened 2016-09-30
Add FantasyLand ChainRec

### LINKED ISSUE #23: Implement FantasyLand ChainRec
https://github.com/fantasyland/fantasy-land/pull/152

Avaq:
Fixes #23 and also makes `.do()` memory- and stack-safe.

I decided to add an abstraction over `chainRec` I called `loop`. It's `chainRec`, but curried and the function is unary. This to me seemed like a more user-friendly version for direct use, since my `done` and `next` construct [Iterations](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#iterator), which should be quite easy for a user to do in user-land. But I would like feedback on this decision. Note that `chainRec` is also available because of Static-Land, but it's not quite as documented.

ping @safareli @rpominov :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,7 +122,7 @@ A list of all types used within the signatures follows:
 - **Apply** - Values which conform to the [Fantasy Land Apply specification][14].
 - **Iterator** - Objects with `next`-methods which conform to the [Iterator protocol][18].
 - **Iteration** - `{done, value}`-Objects as defined by the [Iterator protocol][18].
-- **Next** - An uncomplete (`{done: false}`) Iteration.
+- **Next** - An incomplete (`{done: false}`) Iteration.
 - **Done** - A complete (`{done: true}`) Iteration.
 - **Cancel** - The nullary cancellation functions returned from computations.
 
@@ -266,7 +266,7 @@ a Future of an [Iteration](#type-signatures). If the Iteration is incomplete
 until it returns a Future of a complete (`{done: true}`) Iteration.
 
 For convenience and interoperability, the first two arguments passed to the
-function are functions for creating an uncomplete Iteration, and for creating a
+function are functions for creating an incomplete Iteration, and for creating a
 complete Iteration, respectively.
 
 ```js
--- a/README.md
+++ b/README.md
@@ -62,7 +62,6 @@ getPackageName('package.json')
     * [try](#try)
     * [encase](#encase)
     * [node](#node)
-    * [loop](#loop)
     * [chainRec](#chainrec)
   1. [Transforming Futures](#transforming-futures)
     * [map](#map)
@@ -123,6 +122,8 @@ A list of all types used within the signatures follows:
 - **Apply** - Values which conform to the [Fantasy Land Apply specification][14].
 - **Iterator** - Objects with `next`-methods which conform to the [Iterator protocol][18].
 - **Iteration** - `{done, value}`-Objects as defined by the [Iterator protocol][18].
+- **Next** - An uncomplete (`{done: false}`) Iteration.
+- **Done** - A complete (`{done: true}`) Iteration.
 - **Cancel** - The nullary cancellation functions returned from computations.
 
 ### Creating Futures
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -251,6 +254,30 @@ Future.node(done => fs.readFile('package.json', 'utf8', done))
...
+```js
+const Next = x => ({done: false, value: x});
+const Done = x => ({done: true, value: x});
+Future.loop((x => Future.of(x < 1000000 ? Next(x + 1) : Done(x))), 0)
+.fork(console.error, console.log);
+//> 1000000
+```
+
+#### chainRec
+##### `.chainRec :: ((((b -> Next b), (b -> Done b), b) -> Future a (Iteration b)), b) -> Future a b`
```
Avaq: Maybe I should document `Next` and `Done` under type-signatures, as I've done so far with every named type. But it seems a bit overkill since I'm also documenting `Iteration`, and `Iteration = Next | Done`.

safareli: type of initial value and type of value passed to `Next` constructor should be same, but constructor of `Done` might take value of different type, so signature is not quite correct.

For example:
```js
const initialPage =  { next : 'posts/1' , posts:[]}
const lastPagePostLikes = Future.chainRec(
  (next, done, page) => (page.next
    ? HTTP.fetchJSON(page.next).map(next)
    : Task.of(done(page.posts.map(getLikes).sum()))
  ), initialPage
)
```
here type of accumulator is `Page` but result is `Number`.
and more explicit type of `chainRec` in that case would be:
```hs
chainRec :: ((Page -> Next Page, Number -> Done Number, Page) -> Future r (Iteration Page Number), Page) -> Future r Number
```
where `Iteration a b = Next a | Done b`.

so general `chainRec` should be something like this:
```hs
chainRec :: ((a -> Next a, b -> Done b, a) -> Future r (Iteration a b), a) -> Future r b
```

also as `Iteration`, `Next` and `Done` are just internal thing it could also be like this:
```hs
chainRec :: ((a -> Iteration a b, b -> Iteration a b, a) -> Future r (Iteration a b), a) -> Future r b
```
and then like this
```hs
chainRec :: ((a -> c, b -> c, a) -> Future r c, a) -> Future r b
```

README.md
```diff
@@ -251,6 +254,30 @@ Future.node(done => fs.readFile('package.json', 'utf8', done))
...
+
+Fantasy-Land compliant version of [`loop`](#loop). Passes Next and Done
```
safareli: Fantasy-Land compliant version of 
```diff
-loop
+chainRec
```

fluture.js
```diff
@@ -168,6 +171,35 @@
     if(!isFunction(f)) error$invalidArgument('Future#chain', 0, 'be a function', f);
   }
 
+  function check$loop(f){
+    if(!isFunction(f)) error$invalidArgument('Future.loop', 0, 'be a function', f);
+  }
+
+  function check$chainRec(f){
+    if(!isFunction(f)) error$invalidArgument('Future.chainRec', 0, 'be a function', f);
+    if(!isTernary(f)) error$invalidArgument('Future.chainRec', 0, 'take at least three arguments', f);
```
safareli: I think "at least" is not needed and it should be just "takes three arguments"

test/fluture.test.js
```diff
@@ -459,6 +460,158 @@ describe('Constructors', () => {
...
+    });
+
+    it('throws TypeError when the returned Future does not contain an iteration', () => {
+      const xs = [null, '', {}, {done: true}, {value: 1, done: 1}];
+      const fs = xs.map(x => _ => Future.chainRec((a, b, c) => Future.of((c, x)), 1).fork(noop, noop));
+      fs.forEach(f => expect(f).to.throw(TypeError, /Future.*first call/));
+    });
+
+    it('throws TypeError when the returned Future does not always contain an iteration', () => {
+      const recur = (a, b, i) => i <= 6 ? Future.of(Future.util.Next(i + 1)) : Future.of('hello');
```
safareli: Why don't use `a` instead of `Future.util.Next`?

test/fluture.test.js
```diff
@@ -1404,6 +1580,29 @@ describe('Compliance', function(){
       ));
     });
 
+    describe('ChainRec', () => {
```
safareli: is this duplication of what above is?

rpominov:
Cool idea about using `Iteration` protocol. Not sure though about additional `loop` method.
When writing code programmers will have to answer question "Do I use `loop` or `chainRec`?"; when reading and for example seeing `loop` one may think "I know `chainRec` but what is `loop`, need to check docs". On the other hand APIs are almost identical.

I think both advantages and disadvantages are pretty low, but disadvantages still outweigh.

test/fluture.test.js
Avaq: Oh whoops :zap:

test/fluture.test.js
Avaq: No. It uses the Static-Land compatible functions rather than Fantasy-Land compatible methods inside its law-tests.

README.md
Avaq: Thank you for the clarification. I did not consider the possibility for the Done type to differ, but it makes sense that it's allowed. I shall update the type signature.

README.md
Avaq: This should be "loop", because I intended `loop` to be leading in the eyes of the user (and `chainRec` more like an internal function only exposed for interoperability). Though it seems @rpominov disagrees, I'll need to consider it some more.

fluture.js
Avaq: :zap:

safareli:
Also if one knows `chainRec` from some other Type s/he can reuse the knowledge here too.

Avaq:
I think your arguments are compelling. I will remove `loop`, curry `chainRec` and put it in the spotlights.

README.md
Avaq: What do you think of the following?

```hs
chainRec :: (((b -> Next), (c -> Done), b) -> Future a Iteration) -> b -> Future a c
```

I believe it's correct, and I think it's also the most clear:

* It tells you which callback creates a Next, and which creates a Done.
* It shows that a Future of Iteration must be returned
* It doesn't go into the specifics of the Iteration type, leaving the signature still relatively short and legible.

test/fluture.test.js
```diff
@@ -459,6 +460,140 @@ describe('Constructors', () => {
...
+        while(i < STACKSIZE + 1) yield Future.of(i++);
+        return i;
+      };
+      const m = Future.do(gen);
+      return assertResolved(m, STACKSIZE + 1);
+    });
+
+  });
+
+  describe('.chainRec()', () => {
```
Avaq: Maybe I should also add some tests describing the behavior of `chainRec` when returning it a rejected Future.

safareli: 👍 looks like @rpominov has good tests [here](https://github.com/rpominov/fun-task/blob/master/test/chainRec.js). Also [this](https://github.com/ramda/ramda-fantasy/blob/master/test/future.test.js#L106-L148) is what I have added to ramda-fantasy. You can adopt some of those.

safareli:
> Double the performance of chainRec

Nice 👍

```sh
$ node bench/chainRec.js
Fluture x 637 ops/sec ±2.34% (82 runs sampled)
Ramda Fantasy x 925 ops/sec ±1.63% (81 runs sampled)
FunTask x 321 ops/sec ±2.00% (78 runs sampled)
Fastest is Ramda Fantasy
```

Avaq:
@safareli Mind benchmarking again? I was getting bigger numbers originally, but I made some further improvements :)

safareli:
```
$ node --version
v6.3.0
$ node bench/chainRec.js
Fluture x 1,322 ops/sec ±3.82% (82 runs sampled)
Ramda Fantasy x 980 ops/sec ±1.23% (86 runs sampled)
FunTask x 331 ops/sec ±3.07% (81 runs sampled)
Fastest is Fluture
```

PR #34, opened 2016-10-17
Refactor to classes

Avaq:
This refactors the code from using closure capturing to using class properties. This way of working was inspired by the [FunTask source](https://github.com/rpominov/fun-task/tree/master/src) and has several benefits:

**Performance**

* `chainRec` 50% faster
* `cache` 100% faster
* `race` - 300% faster
* `parallel` - 30% faster
* All other functions - 10-20% faster

**Introspectivity**

Every type of Future can now have its own `toString` function associated with it. Casting a Future to String now shows something much closer to what the user originally typed.
This new introspective nature also allows for the creation of a more informative [sanctuary-def](https://github.com/sanctuary-js/sanctuary-def) type - something for another pull-request.

**Other fixes**

Besides refactoring, I also corrected some minor mis-behaviors:

* The cancellation function of cached Futures no longer unsubscribes all listeners. Instead only the listeners for the specific call to `fork` which created the called `cancel` function will be unsubscribed. If there are other listeners left, the Future will keep running.
* Fix some broken benchmarks.
* When [`Future.or`](/Avaq/Fluture#or), [`Future.race`](/Avaq/Fluture#race), or [`Future.parallel`](/Avaq/Fluture#parallel) settles, all remaining running Futures will be cancelled.
* Now when a [hooked Future](/Avaq/Fluture#hook) is cancelled, its disposal Future is still executed, and immediately cancelled as well. This creates a much better chance for resources to be disposed, and allows a developer to forcefully get rid of resources in a much more logical place: The disposal cancellation function instead of the consumption cancellation function.
* [`Future.recur`](/Avaq/Fluture#recur) was removed in favor of [`Future.chainRec`](/Avaq/Fluture#chainRec)

These changes are backwards-incompatible, and will be released under a new major version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -455,13 +455,27 @@ withConnection(
 .fork(console.error, console.log)
 ```
 
-Be careful when cancelling a hooked Future. If the resource was acquired but not
-yet consumed, it will no longer be disposed. One way to work around this is to
-have the `consume` computation return a cancel function which forcefully
-disposes of the resource.
+In the case that a hooked Future is *cancelled* after the resource was acquired,
+`dispose` will be executed and immediately cancelled. This means that rejections
+which may happen during this disposal are **silently ignored**. To ensure that
+resources are disposed during cancellation, you might synchronously dispose
+resources in the `cancel` function of the disposal Future:
 
-Take care when using this in combination with [`cache`](#cache). Hooking relies
-on the first operation providing a fresh resource every time it's forked.
+```js
+const closeConnection = conn => Future((rej, res) => {
+
+  //We try to dispose gracefully.
+  conn.flushGracefully(err => {
+    if(err) return rej(err);
+    conn.close();
+    res();
+  });
+
+  //On cancel, we force dispose.
+  return () => conn.close();
+
+});
+```
 
 #### finally
 ##### `#finally :: Future a b ~> Future a c -> Future a b`
--- a/README.md
+++ b/README.md
@@ -331,9 +331,9 @@ Future.of(1)
 ```
 
 Note that, due to its lazy nature, the stack and/or heap will slowly fill up as
-you're you chain more and more over the same structure. It's therefore
-recommended that you use [`chainRec`](#chainrec) in cases where you wish to
-`chain` recursively or traverse a large list (10000+ items).
+you chain more over the same structure. It's therefore recommended that you use
+[`chainRec`](#chainrec) in cases where you wish to `chain` recursively or
+traverse a large list (10000+ items).
 
 #### ap
 ##### `#ap :: Future a b ~> Future a (b -> c) -> Future a c`
--- a/README.md
+++ b/README.md
@@ -466,9 +466,12 @@ const closeConnection = conn => Future((rej, res) => {
 
   //We try to dispose gracefully.
   conn.flushGracefully(err => {
-    if(err) return rej(err);
-    conn.close();
-    res();
+    if(err === null){
+      conn.close();
+      res();
+    }else{
+      rej(err);
+    }
   });
 
   //On cancel, we force dispose.
--- a/README.md
+++ b/README.md
@@ -67,7 +67,6 @@ getPackageName('package.json')
     * [map](#map)
     * [bimap](#bimap)
     * [chain](#chain)
-    * [recur](#recur)
     * [ap](#ap)
     * [swap](#swap)
   1. [Error handling](#error-handling)
@@ -331,19 +330,10 @@ Future.of(1)
 //> 2
 ```
 
-#### recur
-##### `#recur :: Future a b ~> (b -> Future a c) -> Future a c`
-##### `.recur :: (b -> Future a c) -> Future a b -> Future a c`
-
-An alternative version of `chain` which does not build up the cancel function.
-This is useful in the case of a never-resolving recursive computation, in
-order to prevent stack-overflow or out-of-memory errors:
-
-```js
-const recursive = () => Future.after(200, 'world').recur(recursive);
-const cancel = recursive();
-process.on('SIGINT', cancel);
-```
+Note that, due to its lazy nature, the stack and/or heap will slowly fill up as
+you're you chain more and more over the same structure. It's therefore
+recommended that you use [`chainRec`](#chainrec) in cases where you wish to
+`chain` recursively or traverse a large list (10000+ items).
 
 #### ap
 ##### `#ap :: Future a b ~> Future a (b -> c) -> Future a c`
--- a/bench/curry.js
+++ b/bench/curry.js
@@ -1,30 +1,55 @@
 const benchmark = require('benchmark');
 const suite = new benchmark.Suite();
 const curry = require('lodash.curry');
+const {util} = require('..');
 
 const lodash = curry(function(a, b){
   return a + b;
 });
 
-const manual = function(a, b){
-  if(arguments.length === 1) return b => manual(a, b);
+function manual(a, b){
+  if(arguments.length === 1) return util.unaryPartial(manual, a);
   return a + b;
-};
+}
+
+function manual3if(a, b, c, d){
+  if(arguments.length === 1) return util.unaryPartial(manual3if, a);
+  if(arguments.length === 2) return util.binaryPartial(manual3if, a, b);
+  if(arguments.length === 3) return util.ternaryPartial(manual3if, a, b, c);
+  return a + b + c + d;
+}
+
+function manual3switch(a, b, c, d){
+  switch(arguments.length){
+    case 1: return util.unaryPartial(manual3switch, a);
+    case 2: return util.binaryPartial(manual3switch, a, b);
+    case 3: return util.ternaryPartial(manual3switch, a, b, c);
+    default: return a + b + c + d;
+  }
+}
 
 suite.add('Lodash uncurried', () => {
   lodash(1, 2);
-});
-
-suite.add('Manual uncurried', () => {
-  manual(1, 2);
 });
 
 suite.add('Lodash curried', () => {
   lodash(1)(2);
 });
 
+suite.add('Manual uncurried', () => {
+  manual(1, 2);
+});
+
 suite.add('Manual curried', () => {
   manual(1)(2);
+});
+
+suite.add('Manual 3 (if)', () => {
+  manual3if(1)(2)(3)(4);
+});
+
+suite.add('Manual 3 (switch)', () => {
+  manual3switch(1)(2)(3)(4);
 });
 
 suite.on('complete', require('./_print'))
--- a/fluture.js
+++ b/fluture.js
@@ -949,7 +949,7 @@
         rej(reason);
       }, function Future$parallel$fork$res(value){
         out[j] = value;
-        ok += 1;
+        ok = ok + 1;
         if(i < _this._length) run(_this._futures[i], i++, c);
         else if(ok === _this._length) res(out);
       });
--- a/fluture.js
+++ b/fluture.js
@@ -449,16 +449,7 @@
 
   function Future$finally(m){
     check$finally(this, m);
-    const _this = this;
-    return new UnsafeFuture(function Future$finally$fork(rej, res){
-      let cancel;
-      const r = _this._f(function Future$finally$rej(e){
-        cancel = m._f(rej, function Future$finally$rej$res(){ rej(e) });
-      }, function Future$finally$res(x){
-        cancel = m._f(rej, function Future$finally$res$res(){ res(x) });
-      });
-      return cancel ? cancel : (cancel = r, function Future$finally$cancel(){ cancel() });
-    });
+    return new FutureFinally(this, m);
   }
 
   function Future$value(f){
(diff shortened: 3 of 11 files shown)


fluture.js
```diff
@@ -132,6 +131,43 @@
...
+  //Creates a dispatcher for a binary method, but takes the object first rather than last.
+  function createInvertedBinaryDispatcher(method){
+    return function invertedBinaryDispatch(m, a, b){
+      if(arguments.length === 1) return unaryPartial(invertedBinaryDispatch, m);
+      if(arguments.length === 2) return binaryPartial(invertedBinaryDispatch, m, a);
+      if(m && typeof m[method] === 'function') return m[method](a, b);
+      error$invalidArgument(`Future.${method}`, 0, `have a "${method}" method`, m);
+    };
+  }
+
```
Avaq: These will eventually disappear as I will depend on `sanctuary-type-classes` for FantasyLand dispatching, rewrite all other dispatchers to just construct their associated classes directly.

README.md
```diff
@@ -465,13 +455,27 @@ withConnection(
...
+which may happen during this disposal are **silently ignored**. To ensure that
+resources are disposed during cancellation, you might synchronously dispose
+resources in the `cancel` function of the disposal Future:
+
+```js
+const closeConnection = conn => Future((rej, res) => {
+
+  //We try to dispose gracefully.
+  conn.flushGracefully(err => {
+    if(err) return rej(err);
```
davidchambers: Are you writing `return rej(err);` to save yourself from typing `{ rej(err); return; }`? If so, I urge you to reconsider as it makes the lambda's return value appear to be significant. I'd find this clearer:

```javascript
if (err == null) {
  conn.close();
  res();
} else {
  rej(err);
}
```

fluture.js
```diff
@@ -41,7 +41,7 @@
   }
 
   function isFuture(m){
-    return (m instanceof FutureClass) || Boolean(m) && m['@@type'] === TYPEOF_FUTURE;
+    return (m instanceof Future) || Boolean(m) && m['@@type'] === TYPEOF_FUTURE;
```
davidchambers: The parens to the left of the `||` are unnecessary. Also, `m != null` would be clearer than `Boolean(m)`.

What's the purpose of the `instanceof` check?

fluture.js
```diff
@@ -376,544 +404,927 @@
...
+    if(arguments.length === 2) return binaryPartial(Future.encase2, f, x);
+    return new FutureEncase(f, x, y);
+  };
+
+  Future.encase3 = function Future$encase3(f, x, y, z){
+    check$encase3(f);
+    if(arguments.length === 1) return unaryPartial(Future.encase3, f);
+    if(arguments.length === 2) return binaryPartial(Future.encase3, f, x);
+    if(arguments.length === 3) return ternaryPartial(Future.encase3, f, x, y);
+    return new FutureEncase(f, x, y, z);
```
davidchambers: Have you considered using `switch` for these?

```javascript
switch (arguments.length) {
  case 1: return unaryPartial(Future.encase3, f);
  case 2: return binaryPartial(Future.encase3, f, x);
  case 3: return ternaryPartial(Future.encase3, f, x, y);
  default: return new FutureEncase(f, x, y, z);
}
```

Also, why do you refer to `Future.encase3` rather than `Future$encase3`? A recursive function is more self-contained if it refers to itself by name rather than via the identifier to which it happens to be bound.

fluture.js
```diff
@@ -376,544 +404,927 @@
...
+  FutureChain.prototype._f = function FutureChain$fork(rej, res){
+    const _this = this;
+    let cancel;
+    const r = _this._parent._f(rej, function FutureChain$fork$res(x){
+      const m = _this._chainer(x);
+      check$chain$f(m, _this._chainer, x);
+      cancel = m._f(rej, res);
     });
-  };
+    return cancel ? cancel : (cancel = r, function FutureChain$fork$cancel(){ cancel() });
```
davidchambers: `cancel ? cancel : ...` is better written `cancel || ...`.

fluture.js
```diff
@@ -376,544 +404,927 @@
...
-        cancel = m._f(rej, res);
-      }, res);
-      return cancel ? cancel : (cancel = r, function Future$chainRej$cancel(){ cancel() });
-    });
+    let cancel = noop, i = 0;
+    (function Future$chainRec$recur(state){
+      let isSync = null;
+      function Future$chainRec$res(it){
+        check$chainRec$it(it, i);
+        i = i + 1;
```
davidchambers: `+=`?

fluture.js
```diff
@@ -376,544 +404,927 @@
...
+        isSync = null;
+        const m = _this._iterate(Next, Done, state.value);
+        check$chainRec$f(m, _this._iterate, i, state.value);
+        cancel = m._f(rej, Future$chainRec$res);
+        if(isSync === true){
+          continue;
+        }else{
+          isSync = false;
+          return;
+        }
```
davidchambers: Is this equivalent to the `if`/`else` above?

```javascript
if (isSync !== true) {
  isSync = false;
  return;
}
```

fluture.js
```diff
@@ -376,544 +404,927 @@
...
+        i = i + 1;
+        if(isSync === null){
+          isSync = true;
+          state = it;
+        }else{
+          Future$chainRec$recur(it);
+        }
+      }
+      while(!state.done){
+        isSync = null;
```
davidchambers: Could we have `isSync :: Boolean` rather than `isSync :: Nullable Boolean`? The latter is confusing and could lead to subtle bugs.

test/fluture.test.js
```diff
@@ -216,6 +249,13 @@ describe('Constructors', () => {
       expect(f).to.throw(error);
     });
 
+    it('has custom toString and inspect', () => {
+      const m = Future.encase((a) => (a), 1);
```
davidchambers: Why `(a) => (a)` rather than `a => a`?

fluture.js
jordalgo: I want to say this is a v8 optimization (shrug).

fluture.js
```diff
@@ -376,544 +404,927 @@
...
+  Future.prototype[FL.bimap] = Future.prototype.bimap;
+  Future.prototype[FL.chain] = Future.prototype.chain;
+  Future.prototype[FL.chainRec] = Future.prototype.chainRec;
+
+  /////////////////
+  // Sub classes //
+  /////////////////
+
+  function SafeFuture(computation){
+    this._computation = computation;
```
jordalgo: raw -> computation 👍

fluture.js
```diff
@@ -376,544 +404,927 @@
...
+    this._queue = [];
+    this._queued = 0;
+    this._value = null;
+    this._state = 0;
   }
 
-  function Future$bimap(f, g){
-    check$bimap(this, f, g);
+  CachedFuture.STATE = {
+    0: 'cold',
```
jordalgo: Why not flip these and have the keys be the names so the code that utilizes the state can be easier to understand?

fluture.js
```diff
@@ -376,544 +404,927 @@
...
     });
-  };
+    return cancel ? cancel : (cancel = r, function FutureChain$fork$cancel(){ cancel() });
+  }
 
-  Future.try = function Future$try(f){
-    return Future.encase(f, undefined);
-  };
+  FutureChain.prototype.toString = function FutureChain$toString(){
+    return `${this._parent.toString()}.chain(${showf(this._chainer)})`;
```
jordalgo: My first reaction to this convention was it's nice to see the parent string stringified but now I'm wondering about long Future chains and what usefulness that monster string would provide.

fluture.js
Avaq: The parens are there because I've often had issues with order of evaluation surrounding `instanceof` checks. I became a bit afraid.
The purpose of the `instanceof` check is to provide a high-performance check for 99% of input to this function, where the right-hand side does not need to be evaluated

fluture.js
Avaq: Referring to `Future.encase3` is a legacy thing. I changed most during refactoring but I may have missed a few. Good catch. :)
Good idea about the switch statement. I'll have to see which performs faster.

fluture.js
Avaq: `isSync` has three states; `undetermined`, `true` and `false`. Perhaps I could represent them with numbers.

test/fluture.test.js
Avaq: Copy-pasting

fluture.js
davidchambers: Numbers would be an improvement. This would be even clearer:

```javascript
//    data Timing = Synchronous | Asynchronous | Undetermined

//    Synchronous :: Timing
const Synchronous = 'Synchronous';

//    Asynchronous :: Timing
const Asynchronous = 'Asynchronous';

//    Undetermined :: Timing
const Undetermined = 'Undetermined';
```

fluture.js
Avaq: Actually; I'm also asserting `state > 1` here and there. Assigning numbers to the state encodes order, which I can then use to determine how far through its state a CachedFuture is in one simple operation.

fluture.js
```diff
@@ -376,544 +404,927 @@
...
-      });
-      const c2 = m._f(rej, function Future$ap$resThat(f){
-        if(!ok2) return void (ok1 = 1, _f = f);
-        check$ap$f(f);
-        res(f(_x));
-      });
-      return function Future$ap$cancel(){ c1(); c2() };
-    });
+  CachedFuture.prototype._drainQueue = function CachedFuture$drainQueue(){
+    const q = this._queue, l = q.length, s = this._state, v = this._value;
```
Avaq: No need  to name these such small names. I actually just did this to move property-access to outside of the loop.

fluture.js
Avaq: `i = i + 1`:
```sh
$ node bench/chainRec.js 
Fluture x 1,577 ops/sec ±0.33% (93 runs sampled)
```

`i += 1`:
```sh
$ node bench/chainRec.js 
Fluture x 1,249 ops/sec ±0.33% (92 runs sampled)
```

Avaq:
Think this is ready to merge @davidchambers @jordalgo ? :)

bench/curry.js
```diff
@@ -1,31 +1,56 @@
...
-const manual = function(a, b){
-  if(arguments.length === 1) return b => manual(a, b);
+function manual(a, b){
+  if(arguments.length === 1) return util.unaryPartial(manual, a);
   return a + b;
-};
+}
+
+function manual3if(a, b, c, d){
+  if(arguments.length === 1) return util.unaryPartial(manual3if, a);
```
jordalgo: This obviously can wait for another PR/commit, but make this switch just to be consistent?

Avaq: No. The idea here is to test the performance of `if` vs `switch`. That's why the function is called manual3**if** ^^

fluture.js
```diff
@@ -376,544 +404,950 @@
...
+  }
+
+  FutureTry.prototype = Object.create(Future.prototype);
+
+  FutureTry.prototype._f = function FutureTry$0$fork(rej, res){
+    let r;
+    try{ r = this._fn() }catch(e){ rej(e); return noop }
+    res(r);
+    return noop;
+  }
```
Avaq: We can't have the call to `res` reside inside the `try` block, as it will cause any subsequent actions triggered by calling `res` to also have their thrown Errors caught, causing the Future to enter the rejection branch after already being resolved... [scary stuff](https://github.com/Avaq/Fluture/commit/e9006a8606dc9f635c564a7d570229eb4c4255df).

jordalgo: Ah right! Maybe I'll add a unit test to try for this also. 👍

jordalgo:
@Avaq This is looking good. Nice work!

fluture.js
davidchambers: I'm very surprised to learn that the two are not identical. I expected `i += 1` to be rewritten to `i = i + 1` prior to evaluation, but apparently this canonicalization does not occur.

davidchambers:
I :heart: the named states. The changes look great to me!

Avaq:
@jordalgo I've tagged a release candidate. You can install it with `npm i -S avaq/fluture#4.0.0-rc.1`. I find that in my existing projects, the `toString` seems very reasonable, with long chains of Futures typically being around 100 lines when cast to String. Since these strings really only appear in error messages, and are very helpful to determine which Future is saying the error, I think it's totally worth it.

jordalgo:
@Avaq Pulled it down and it looks good. The `toString` method is actually pretty nice when I tried invoking it directly on a long chain 👍 However (and just from a few minutes of tinkering) I wasn't able to get a long toString error message when passing some bad args to methods in the chain. Do you have a sample of when this gets used in the case of an error?

Avaq:
```js
const m = Future.of() //.long().chain()
Future(m)
```

jordalgo:
Ok. Looks good, though still not sure how often it will get used :) This version looks good 👍

Avaq:
It's been released; https://github.com/Avaq/Fluture/releases/tag/4.0.0

PR #35, opened 2016-10-20
Replace Future#promise() with Promises/A+ compatible Future#then()

Avaq:
So this might be an interesting topic for conversation. Using the new `CachedFuture` class I created while working on #34, I was able to create a fully Promises/A+ compliant `then` method. Essentially **turning Fluture into a Promise library**! o.0

I think the library might benefit from having this. Particularly because `fork` is not standardized. Though I'm very eager to hear peoples opinions. Check out the diff in the readme to learn about how I advice the feature to be used.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,7 +79,7 @@ getPackageName('package.json')
   1. [Consuming Futures](#consuming-futures)
     * [fork](#fork)
     * [value](#value)
-    * [promise](#promise)
+    * [then](#then)
   1. [Parallelism](#parallelism)
     * [race](#race)
     * [or](#or)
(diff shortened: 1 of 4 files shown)


Avaq:
An alternative is to provide `then` as such:

```js
(onReject, onResolve) => {
  return new Promise((res, rej) => this._f(rej, res)).then(onReject, onResolve);
}
```

And not concern myself with the complexity of Promises. The intended use for `then` would remain the same. The only difference being that you won't be able to use Future methods on the return value - perhaps even for the best.

davidchambers:
I'm concerned that this may confuse new users of the library.

Avaq:
I think it's quite cool that Futures can also be Promises. But that's where I run out of pro's. I think making Futures Promises enables a bunch of bad practices and nothing more. I think providing a `then` function which returns a Promise has less pitfalls, but provides nothing more than a magical convenience over `.promise()`. I think this PR will enjoy its new life as reference material. ;)

PR #37, opened 2016-10-22
Add extractLeft and extractRight methods and dispatchers

Avaq:
Mainly for use with Sanctuary-Def, as documented. What do you think, @davidchambers?

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,7 @@ getPackageName('package.json')
     * [isForkable](#isforkable)
     * [cache](#cache)
     * [do](#do)
+  1. [Sanctuary](#sanctuary)
   1. [Futurization](#futurization)
 - [Benchmarks](#benchmarks)
 - [The name](#the-name)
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -781,6 +782,40 @@ Future.do(function*(){
...
+);
+
+//...
+
+const toErrorMessage = def(
+  'toErrorMessage',
+  {},
+  [FutureType($.Error, a), FutureType($.Error, a)],
+  Future.mapRej(e => e.message)
+);
```
davidchambers: I'm confused by this function. What is it meant to demonstrate, exactly? Also, should the return type be `FutureType($.String, a)` rather than `FutureType($.Error, a)`?

davidchambers:
I assume you don't want to export `FutureType` as you don't want to add sanctuary-def as a dependency due to its size. This suggests to be that the type constructors (such as `$.BinaryType`) should be defined in a separate package so one can define Sanctuary types without depending on sanctuary-def.

Do you think `extractLeft` and `extractRight` are useful in *other* contexts? If not, you could consider providing a function for *creating* `FutureType :: (Type, Type) -> Type`. It might look like this:

```javascript
//    createFutureType :: ((String, Any -> Boolean, Future a b -> Array a, Future a b -> Array b) -> (Type, Type) -> Type) -> (Type, Type) -> Type
const createFutureType = BinaryType => BinaryType(
  Future.name,
  Future.isFuture,
  Future._extractLeft,
  Future._extractRight
);
```

Usage example:

```javascript
const $ = require('sanctuary-def');

//    FutureType :: (Type, Type) -> Type
const FutureType = Future.createFutureType($.BinaryType);
```

That's a scary type signature, though!

Avaq:
> [Y]ou could consider providing a function for creating `FutureType`

I considered exactly the same function, but opted for documenting `extractLeft` and `extractRight` like this, due to the complexity of `createFutureType` and difficulty to ensure valid input.

davidchambers:
:deciduous_tree:

If and when we move `$.BinaryType` and friends to a separate package we can consider providing `FutureType :: Type` directly.

PR #43, opened 2016-11-04
sanctuary-type-classes@^1.0.0

davidchambers:
https://github.com/sanctuary-js/sanctuary-type-classes/compare/v0.3.0...v1.0.0

v1.0.0 is equivalent to v0.3.1. I released the former upon your suggestion, @Avaq. The library is solid.


Avaq:
Thanks! :deciduous_tree:

PR #55, opened 2016-12-30
Utilize gihoma to set up git hooks

mickvangelderen:
`npm run setup` used to copy anything from `scripts/hooks` to
`.git/hooks`. This has a number of drawbacks:

1. Non-executable files are also copied.
2. If a script is updated, every clone needs to run `npm run setup`
again.
3. Scripts can not have a file extension.

The `gihoma` module solves all of these issues.

Existing clones will have to run
`node_modules/.bin/gihoma install --verbose --force` to overwrite
existing files in their .git/hooks folder with symlinks.


Avaq:
This is a great idea. Though I recently [learned about Husky](https://github.com/Avaq/node-server-skeleton/pull/14), which appears to be more convenient. It installs the git hooks as part of its `npm install`. So it removes the need for a `setup` command altogether.

mickvangelderen:
Ah, `husky` looks very convenient and if that works for this repository that is fine. 

My personal preference to use as little 'magic' as possible, and opting in to git hooks which are there for the developers' convenience.

PR #58, opened 2017-01-03
Add es5 support

lithin:
## Problem
When doing uglification for our production app, we ran into problems with UglifyJS2 not supporting ES6.

## Solution
Normally, npm packages are transpiled when published so that users of the packages don't have to worry about their compatibility. I thought it might be best to give users the choice - by default they import ES6 version but optionally can `require('Fluture/es5')` to avoid issues when compiling for production.

Same thing will need to be done for `inspect-f` as it's a dependency for Fluture. In which case, Fluture will need to require ES5 version of `inspect-f`.

Do you think this would be a good way to go?

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,14 +3,19 @@
   "version": "4.2.0",
   "description": "FantasyLand compliant (monadic) alternative to Promises",
   "main": "fluture.js",
+  "files": [
+    "es5.js"
+  ],
   "repository": "https://github.com/Avaq/fluture.git",
   "scripts": {
+    "build": "babel fluture.js -o es5.js",
     "clean": "rimraf npm-debug.log coverage",
     "lint": "eslint fluture.js test",
     "lint:readme": "remark --no-stdout --frail -u remark-validate-links README.md",
     "postcheckout": "npm install",
     "postmerge": "npm install",
     "precommit": "npm run lint && npm run lint:readme && npm run test:unit",
+    "prepublish": "npm run build",
     "release": "npm outdated --long && xyz --edit --repo git@github.com:Avaq/Fluture.git --tag 'X.Y.Z' --increment",
     "toc": "node scripts/toc.js",
     "test": "npm run test:all && codecov",
@@ -29,9 +34,6 @@
   "engines": {
     "node": ">=4.0.0"
   },
-  "files": [
-    "fluture.js"
-  ],
   "keywords": [
     "algebraic",
     "async",
@@ -52,10 +54,12 @@
     "sequential"
   ],
   "dependencies": {
-    "inspect-f": "^1.1.0",
+    "inspect-f": "^1.2.0",
     "sanctuary-type-classes": "^2.0.1"
   },
   "devDependencies": {
+    "babel-cli": "^6.18.0",
+    "babel-preset-latest": "^6.16.0",
     "benchmark": "^2.1.0",
     "chai": "^3.4.0",
     "codecov": "^1.0.1",
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,3 @@
+{
+  "presets": ["latest"]
+}
--- a/.gitignore
+++ b/.gitignore
@@ -28,5 +28,8 @@ coverage/
 # Test file - who doesn't have a test file?
 test.js
 
+# Transpiled ES5 file - for publishing only
+es5.js
+
 # Heap snapshots
 *.heapsnapshot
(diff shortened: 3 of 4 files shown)


Avaq:
Hi @lithin! Thank you for the PR. It seems you've amended the latest commit on `master`, rather than creating a commit on top of it. Would you mind doing:

```sh
git reset --soft d9e7fffc386da872c7a47b4d9b6deb05424550c8
git commit
git push --force
```

package.json
```diff
@@ -3,14 +3,19 @@
   "version": "4.2.0",
   "description": "FantasyLand compliant (monadic) alternative to Promises",
   "main": "fluture.js",
+  "files": [
+    "es5.js"
+  ],
```
Avaq: Shouldn't `fluture.js` remain in the publishing files?

lithin: `fluture.js` is automatically published as it is the `main` entry point - so there's no need :)

lithin: I can move it back in though if you prefer

Avaq:
> Do you think this would be a good way to go?

I think it offers a convenience I've would like to offer. But just to get an idea of your use-case; since you are doing uglification, wouldn't you be able to simply include a build-step prior to this which transpiles ES6 down to ES5?

package.json
Avaq: I was unaware. Let's leave it out. :+1:

lithin:
> ...wouldn't you be able to simply include a build-step prior to this which transpiles ES6 down to ES5?

Yes, we transpile our code but we exclude `node_modules`. That's mainly because that would make the build take really long - but also it's not required by any other modules than `Fluture` and `inspect-f`.

Just out of interest - do you know of any apps that use `Fluture` in prod?

Avaq:
> it's not required by any other modules than `Fluture` and `inspect-f`.

I guess that's a convincing argument. 'Would not want to be that guy. Are you working on a similar PR for `inspect-f`? I think it may be better to just rewrite `inspect-f` to use ES5-only. It's only a few lines.

> do you know of any apps that use Fluture in prod?

I know of quite a few, but I'm afraid none are open-source. The company I work at has adopted using https://github.com/Avaq/node-server-skeleton for some of its recent projects. I think this code comes closest to seeing how Fluture is used in production. I also know that @jgoux has quite some experience using Fluture in production, maybe he can provide some examples.

-----

PS: I saw the talk about Fluture from your colleague. It came as a pleasant surprise that people are giving talks about it now! :grinning:

Avaq:
> I think it may be better to just rewrite inspect-f to use ES5-only.

If you'll agree then I will get on this and update the development tools as I did yesterday for Fluture. I expect to have an ES5 compatible version out in some hours.

lithin:
@Avaq If you can rewrite `inspect-f`, that would be perfect!

Ahh I guess people are using it mainly on back-end so far so they don't need to sort out cross-browser issues? Thanks for the link, I think we all needed some real-life examples :)

I didn't even know someone did a talk on Fluture - I'll check it out!

Avaq:
> I guess people are using it mainly on back-end so far

That's right.

Avaq:
You mind taking a look at https://github.com/Avaq/inspect-f/pull/2 ?

Avaq:
### Some things to do before merging

* [ ] Merge https://github.com/Avaq/inspect-f/pull/2
* [ ] Release a new minor `inspect-f` version
* [ ] Update `package.json` in this branch to have the latest `inspect-f` version as a dependency
* [ ] Remove the "requires node 4.0" notice from the README
* [ ] Document under **usage** that `require('fluture/es5')` is an option

lithin:
I've done a quick review on `inspect-f` and updated readme for `Fluture` so the last two boxes are checked :)

Avaq:
I released `inspect-f@1.2.0`. You can add it to the `package.json`.

lithin:
Done 👍

Avaq:
Would you mind squashing your commits?

```sh
git reset --soft 027447d50ab6c7103b1e3c1ac1f5c867bfb3598a
git commit --amend
git push --force
```

And I'm not sure if it is possible, but perhaps you could depend on this branch in the project which prompted the issue to see if ES5 is now properly supported.

lithin:
It works!! :) Thanks so much for your help. I think it's good to go.

Avaq:
Your changes were published in version `4.3.0` :)
Thank you for your contributions. I hope you'll enjoy using the lib!

PR #59, opened 2017-01-05
Integrate sanctuary-type-identifiers

Avaq:
This might be considered a breaking change, since it's a feature intended for interop with Sanctuary, and I believe it will break all `S.is(Future, x)` calls.


davidchambers:
:deciduous_tree:

PR #65, opened 2017-01-28
Remove husky to let developer decide whether to use git hooks

Avaq:
Better like this @davidchambers ? :)


davidchambers:
:deciduous_tree:

Issue #44, opened 2016-11-07
Expose non monadic version of Future


safareli:
It will be nice to expose non monadic version of Future for which `ap` would be parallel.
This version could also implement [Alternative](https://github.com/fantasyland/fantasy-land/pull/197) where `alt` would be `race`

Avaq:
:information_source: I'm working on this.

Avaq:
I'm looking into developing this as its own package, extending Fluture with an optional Parallel structure. Some of Flutures core functionality will also be separated out to serve as a base for both packages.

safareli:
@Avaq why?

So you are planning to have 3 repos like this?:
- Avaq/Fluture-core
- Avaq/Fluture
- Avaq/FluturePar

Avaq:
> why?

In writing a detailed response about my reasons, I realized there may be some issues with this approach (circular dependencies, shared code). I'm going to drop the idea, and keep it in the same package.

safareli:
After discussing this offline, here is what we came to:

As there is [Concurrently](https://hackage.haskell.org/package/async-2.1.1/docs/Control-Concurrent-Async.html) in haskell:
```hs
newtype Concurrently a = Concurrently { runConcurrently :: IO a }

instance Functor Concurrently where
  fmap f (Concurrently a) = Concurrently $ f <$> a

instance Applicative Concurrently where
  pure = Concurrently . return
  Concurrently fs <*> Concurrently as =
    Concurrently $ (\(f, a) -> f a) <$> concurrently fs as

instance Alternative Concurrently where
  empty = Concurrently $ forever (threadDelay maxBound)
  Concurrently as <|> Concurrently bs =
    Concurrently $ either id id <$> race as bs
```

We could have something like this:
```js
const mkConcurrently = (TaskRep, concurrently, race, zero) => {
  const Concurrently = daggy.tagged('run')
  Concurrently[fl.zero] = zero
  Concurrently[fl.of] = (a) => Concurrently(TaskRep[fl.of](a))
  Object.assign(Concurrently.prototype, {
    [fl.map](f) {
      return Concurrently(this.run[fl.map](f))
    },
    [fl.ap](a) {
      return Concurrently(concurrently(this.run,a))
    },
    [fl.alt](a) {
      return Concurrently(race(this.run, a))
    },
  })
  return Concurrently
}
```

which could be used with any Task/Future/whatever.

example with Fluture:
```js
const Par  =  mkConcurrently(
  Fluture, 
[...]

PR #67, opened 2017-01-31
Deprecate fromForkable and isForkable (again)

Avaq:
I deprecated `cast` in version 4.2, but reinstated it as `fromForkable` in version 4.3 because I had plans to create a second "forkable" type; `Parallel`. The idea was that you would be able to cast back and forth between the two like `Future.fromForkable(Parallel.of(1))` and `Parallel.fromForkable(Future.of(1))`. After some discussion in #44, the API was changed to `par = Parallel(Future.of(1))` and (probably) `Future.seq(par)`.

Since in the new setup, `Parallel` is not a "forkable" and Future remains the only local "forkable", I would like to deprecate `fromForkable` as well, for the same reasons I deprecated `cast` in the first place.

The functions are to be removed in the next major release, and deprecation gives a few weeks head to developers. I expect the usage of the functions to be very limited.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,6 @@ For front-end applications and node <v4, please use `require('fluture/es5')`.
     * [rejectAfter](#rejectafter)
     * [try](#try)
     * [encase](#encase)
-    * [fromForkable](#fromforkable)
     * [fromPromise](#frompromise)
     * [node](#node)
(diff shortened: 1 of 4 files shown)


Avaq:
I'm going to assume that I'm right in thinking that nobody uses this.

PR #68, opened 2017-01-31
Remove custom inspect functions

Avaq:
After hearing out @safareli's [arguments](https://github.com/fluture-js/concurrify/pull/1#discussion_r98405015) about custom `inspect` for [concurrify](https://github.com/fluture-js/concurrify), I went along and checked if the standard inspect provides useful output, and I would argue that it does:

```js
> Future.after(1, 1).map(x => x + 1).chain(x => Future.of(x)).toString()
'Future.after(1, 1).map(x => x + 1).chain(x => Future.of(x))'

> Future.after(1, 1).map(x => x + 1).chain(x => Future.of(x))
Future {
  _parent: Future { _parent: Future { _time: 1, _value: 1 }, _mapper: [Function] },
  _chainer: [Function] }
```

Reasons for removing custom inspect:

* Default inspect lies within the expectation of a developer
* Default inspect may improve in the future
* `toString` might overflow the stack, potentially giving developers a very bad REPL experience
* Why provide one way to inspect a Future if we can provide two?

---

Should this be considered a breaking change? I would say "no" because `m.inspect` is still `Future ab ~> () -> String` and the string it returns is undocumented and should not be built upon.

### DIFF
--- a/fluture.js
+++ b/fluture.js
@@ -251,10 +251,6 @@
     return this._f(rej, res);
   };
 
-  Future.prototype.inspect = function Future$inspect(){
-    return this.toString();
-  };
-
   Future.prototype.value = function Future$value(f){
     if(!isFuture(this)) invalidContext('Future#value', this);
     if(!isFunction(f)) invalidArgument('Future#value', 0, 'be a function', f);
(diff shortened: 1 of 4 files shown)


davidchambers:
> Should this be considered a breaking change? I would say "no" because `m.inspect` is still `Future ab ~> () -> String` and the string it returns is undocumented and should not be built upon.

I agree.

safareli:
just saw this: http://code.visualstudio.com/updates/v1_9#_better-object-display-in-the-console-node2

Avaq:
Oh. So we made a good choice? :)

safareli:
yap 🥇

PR #66, opened 2017-01-30
Add fast failing to all remaining curried functions

Avaq:
> Finally closes #4 and fixes a bug where Future.bimap would throw
> a TypeError when given exactly two valid arguments.

PR #69, opened 2017-02-10
Add fast failing to all remaining curried functions

### LINKED ISSUE #4: Fail-fast curried functions
Fail as soon as the invalid argument is provided to the curried function.
- [x] ap
- [x] map
- [x] bimap
- [x] chain
- [x] and
- [x] rejectAfter
- [x] chainRej
- [x] mapRej
- [x] swap
- [x] fork
- [x] race
- [x] or
- [x] fold
- [x] hook
- [x] finally
- [x] after
- [x] cast
- [x] encase
- [x] encase2
- [x] encase3
- [x] parallel
- [x] value

Avaq:
> Finally finally closes #4

I've reverted the commits involved in #66 because it broke multiple of the affected functions. The reason some of the functions broke is because of arguments being passed into class constructors in the wrong order. The reason this wasn't caught by unit-tests is because all I'm asserting for the static constructors, is that they return the expected class instance. The actual test of the class instances themselves use instances created by methods (rather than the static functions affected in this commit).

Now I'm reluctant to merge this until I know that these kind of mistakes would be caught by the unit tests. Maybe I should run all instance tests on several instances, each created in a different way. For now that means that most instance tests would run twice.

### DIFF
--- a/fluture.js
+++ b/fluture.js
@@ -114,52 +114,15 @@
     return function partial(d){ return f(a, b, c, d) };
   }
 
-  //Creates a dispatcher for a nullary method.
-  function createNullaryDispatcher(method){
-    return function nullaryDispatch(m){
-      if(m && typeof m[method] === 'function') return m[method]();
-      return error$invalidArgument(`Future.${method}`, 1, `have a "${method}" method`, m);
-    };
-  }
-
-  //Creates a dispatcher for a unary method.
-  function createUnaryDispatcher(method){
-    return function unaryDispatch(a, m){
-      if(arguments.length === 1) return unaryPartial(unaryDispatch, a);
-      if(m && typeof m[method] === 'function') return m[method](a);
-      return error$invalidArgument(`Future.${method}`, 1, `have a "${method}" method`, m);
-    };
-  }
-
-  //Creates a dispatcher for a binary method.
-  function createBinaryDispatcher(method){
-    return function binaryDispatch(a, b, m){
-      if(arguments.length === 1) return unaryPartial(binaryDispatch, a);
-      if(arguments.length === 2) return binaryPartial(binaryDispatch, a, b);
-      if(m && typeof m[method] === 'function') return m[method](a, b);
-      return error$invalidArgument(`Future.${method}`, 2, `have a "${method}" method`, m);
-    };
-  }
-
-  //Creates a dispatcher for a binary method, but takes the object first rather than last.
-  function createInvertedBinaryDispatcher(method){
-    return function invertedBinaryDispatch(m, a, b){
-      if(arguments.length === 1) return unaryPartial(invertedBinaryDispatch, m);
-      if(arguments.length === 2) return binaryPartial(invertedBinaryDispatch, m, a);
-      if(m && typeof m[method] === 'function') return m[method](a, b);
-      return error$invalidArgument(`Future.${method}`, 0, `have a "${method}" method`, m);
-    };
-  }
-
   //Creates an error about an invalid argument.
-  function error$invalidArgument(it, at, expected, actual){
+  function invalidArgument(it, at, expected, actual){
     throw new TypeError(
       `${it} expects its ${ordinal[at]} argument to ${expected}\n  Actual: ${show(actual)}`
     );
   }
 
   //Creates an error message about a method being called with an invalid context.
-  function error$invalidContext(it, actual){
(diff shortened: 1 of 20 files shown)


fluture.js
```diff
@@ -479,46 +442,178 @@
...
+
+  Future.chainRej = function chainRej(chainer, m){
+    if(!isFunction(chainer)) invalidArgument('Future.chainRej', 0, 'be a Function', chainer);
+    if(arguments.length === 1) return unaryPartial(chainRej$chainer, chainer);
+    return chainRej$chainer(chainer, m);
+  };
+
+  function mapRej$mapper(mapper, m){
+    if(!isFuture(m)) invalidArgument('Future.mapRej', 1, 'be a Future', m);
+    return new FutureMapRej(m, mapper);
```
Avaq: In #66, this line incorrectly contained `return new FutureMapRej(mapper, m);`

fluture.js
```diff
@@ -479,46 +442,178 @@
...
+  Future.fork = function fork(f, g, m){
+    if(!isFunction(f)) invalidArgument('Future.fork', 0, 'be a function', f);
+    if(arguments.length === 1) return unaryPartial(fork$f, f);
+    if(arguments.length === 2) return fork$f(f, g);
+    return fork$f(f, g, m);
+  };
+
+  function fold$f$g(f, g, m){
+    if(!isFuture(m)) invalidArgument('Future.fold', 2, 'be a Future', m);
+    return new FutureFold(f, g, m);
```
Avaq: This line is also incorrect. The first argument should be `m`.

fluture.js
```diff
@@ -479,46 +442,178 @@
...
+
+  Future.or = function or(right, left){
+    if(!isFuture(right)) invalidArgument('Future.or', 0, 'be a Future', right);
+    if(arguments.length === 1) return unaryPartial(or$right, right);
+    return or$right(right, left);
+  };
+
+  function finally$left(left, right){
+    if(!isFuture(right)) invalidArgument('Future.finally', 1, 'be a Future', right);
+    return new FutureFinally(left, right);
```
Avaq: These arguments were also flipped compared to the previous version, which was a far less obvious breaking change.

test/5.future-and.test.js
```diff
@@ -2,113 +2,65 @@
 
 const expect = require('chai').expect;
 const Future = require('../fluture.js');
-const FutureAnd = Future.classes.FutureAnd;
```
Avaq: I prefer this new way of testing, because it tests behavior rather than implementation.

test/5.future-and.test.js
```diff
@@ -117,11 +69,51 @@ describe('FutureAnd', () => {
...
+    expect(f).to.throw(TypeError, /Future.*first/);
+  });
+
+  it('throws when not given a Future as second argument', () => {
+    const f = () => Future.and(Future.of(1), 1);
+    expect(f).to.throw(TypeError, /Future.*second/);
+  });
+
+  //TODO: Argument order.
+  testInstance((a, b) => Future.and(a, b));
```
Avaq: `Future.and` was one of the functions I implemented without a `createDispatcher` from the start, and I made the same mistake there that caused all this mess now. Essentially the argument order of `Future.and` is the wrong way around (static functions should be flipped). If I'd fix that now, it would be a breaking change, hence the `#TODO`.

test/5.future-both.test.js
```diff
@@ -117,11 +69,51 @@ describe('FutureBoth', () => {
...
+    expect(f).to.throw(TypeError, /Future.*first/);
+  });
+
+  it('throws when not given a Future as second argument', () => {
+    const f = () => Future.both(Future.of(1), 1);
+    expect(f).to.throw(TypeError, /Future.*second/);
+  });
+
+  //TODO: Argument order.
+  testInstance((a, b) => Future.both(a, b));
```
Avaq: See my comment on `future-and.test.js`.

Avaq:
Functions that turned out to be broken after fixing the tests:

* `mapRej`
* `chainRej`
* `race`
* `finally`
* `fold`

test/5.future-both.test.js
Avaq: Actually, it's quite expectable that `Future.both(ma, mb)` would give me `Future [a ,b]` and not `Future [b, a]`. I'm going to remove this comment.

PR #70, opened 2017-02-13
Add ConcurrentFuture based on concurrify

Avaq:
Adds the following functions:

* `Future.Par :: Future a b -> ConcurrentFuture a b`
* `Future.seq :: ConcurrentFuture a b -> Future a b`
* `Future.never :: () -> Future a a`

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,7 @@ For front-end applications and node <v4, please use `require('fluture/es5')`.
   1. [Creating Futures](#creating-futures)
     * [Future](#future)
     * [of](#of)
+    * [never](#never)
     * [reject](#reject)
     * [after](#after)
     * [rejectAfter](#rejectafter)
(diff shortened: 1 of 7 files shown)


Avaq:
Oh. Forgot documentation again. On its way.

Avaq:
I think that's it. Would you mind taking a look @safareli?

README.md
```diff
@@ -172,6 +174,12 @@ eventualThing.fork(
 //> "Hello world!"
 ```
 
+#### never
+##### `.never :: () -> Future a a`
```
Avaq: I'll change this to `never :: Future a a`.

fluture.js
```diff
@@ -1565,6 +1606,22 @@
...
+
+  FutureNever.prototype._f = function FutureNever$fork(){
+    return noop;
+  };
+
+  FutureNever.prototype.toString = function FutureNever$toString(){
+    return `Future.never`;
+  };
+
+  Future.never = new FutureNever;
```
safareli: btw is there some reason of having separate class `FutureNever` when it could be implemented like this (as there is only one instance):

```js
Future.never = object.assign(Object.create(Future.prototype),{
  _f: function FutureNever$fork(){
    return noop;
  },
  toString: function FutureNever$toString(){
    return `Future.never`;
  },
})
```

Avaq: There's no reason for it. But I guess it's not really important. I think it's fine like this.

safareli: it's fine

PR #76, opened 2017-02-15
Improve documentation on interoperability

Avaq:
[HTML Preview](https://github.com/fluture-js/Fluture/blob/av-improve-interoparability-docs/README.md#interoperability)

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,12 +102,13 @@ For front-end applications and node <v4, please use `require('fluture/es5')`.
 [<img src="https://raw.github.com/fantasyland/fantasy-land/master/logo.png" align="right" width="82" height="82" alt="Fantasy Land" />][FL]
 [<img src="https://raw.githubusercontent.com/rpominov/static-land/master/logo/logo.png" align="right" height="82" alt="Static Land" />][6]
 
-Fluture implements [FantasyLand 1][FL1], [FantasyLand 2][FL2],
-[FantasyLand 3][FL3] and [Static Land][6] compatible `Functor`, `Bifunctor`,
-`Apply`, `Applicative`, `Chain`, `ChainRec` and `Monad`.
-
-FantasyLand 0.x is *mostly* supported. Everything but `Apply` (`ap`) is, this
-means that dispatchers to FantasyLand 0.x `ap` will not work.
+* `Future` implements [Fantasy Land 1][FL1], [Fantasy Land 2][FL2],
+  [Fantasy Land 3][FL3], and [Static Land][6] -compatible `Bifunctor`, `Monad`
+  and `ChainRec` (`of`, `ap`, `map`, `bimap`, `chain`, `chainRec`). Fantasy Land
+  0.x is *mostly* supported. Everything but `Apply` (`ap`) is, this means that
+  dispatchers to Fantasy Land 0.x `ap` (like the one in Ramda) will not work.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -102,12 +102,13 @@ For front-end applications and node <v4, please use `require('fluture/es5')`.
...
 [<img src="https://raw.githubusercontent.com/rpominov/static-land/master/logo/logo.png" align="right" height="82" alt="Static Land" />][6]
 
-Fluture implements [FantasyLand 1][FL1], [FantasyLand 2][FL2],
-[FantasyLand 3][FL3] and [Static Land][6] compatible `Functor`, `Bifunctor`,
-`Apply`, `Applicative`, `Chain`, `ChainRec` and `Monad`.
-
-FantasyLand 0.x is *mostly* supported. Everything but `Apply` (`ap`) is, this
-means that dispatchers to FantasyLand 0.x `ap` will not work.
+* `Future` implements [FantasyLand 1][FL1], [FantasyLand 2][FL2],
+  [FantasyLand 3][FL3] and [Static Land][6] compatible `Bifunctor`, `Monad` and
```
davidchambers: Suggestions:

  - consider a comma after "FantasyLand 3";
  - consider a hyphen before "compatible" (i.e. _and Static Land -compatible_); and
  - consider writing Fantasy Land as two words (as it is written in the [spec][1]).


[1]: https://github.com/fantasyland/fantasy-land

README.md
```diff
@@ -102,12 +102,13 @@ For front-end applications and node <v4, please use `require('fluture/es5')`.
...
+  `ChainRec` (`of`, `ap`, `map`, `bimap`, `chain`, `chainRec`). FantasyLand 0.x
+  is *mostly* supported. Everything but `Apply` (`ap`) is, this means that
+  dispatchers to FantasyLand 0.x `ap` will not work.
```
davidchambers: Perhaps we should mention `R.ap` here. That's undoubtedly the most popular function which relies on the old `ap` type.

README.md
```diff
@@ -102,12 +102,13 @@ For front-end applications and node <v4, please use `require('fluture/es5')`.
...
+* `Future.Par` implements [FantasyLand 3][FL3] `Alternative` (`of`, `zero`, `map`, `ap`, `alt`).
+* Instances can be identified by, and are compared using, [Sanctuary Type Identifiers][STI].
```
davidchambers: What does it mean to "identify" a future?

What is meant by "compared"? Are you saying that a future is not considered equal to a non-future because the type identifiers do not match?

README.md

README.md
Avaq: Sure. It's the reason I now only mention Sanctuary in the Fluture introduction. I used to also mention Ramda.

README.md
Avaq: > What does it mean to "identify" a future?

I guess technically I mean that `type()` from `sanctuary-type-identifiers` will return an identifier unique to Futures when called in Future instances.

> What is meant by "compared"?

I mean that I use `type()` from `sanctuary-type-identifiers` internally to determine that values I'm being passed are members of Future.

It seems as this whole sentence may need to be rephrased.

Avaq: The goal, really, is just to indicate that the Sanctuary Type Identifiers spec is being followed.

Avaq: Hey! This discussion is not "outdated"!

PR #80, opened 2017-03-31
ES6 Rework

Avaq:
This painfully Object Oriented (*see 6*) proof of concept represents the beginning of rewriting Fluture using es6 modules, with some goals in mind:

1. **Support tree shaking** - The more code I can separate to independent chunks, the smaller `rollup` can make bundles which include Fluture.
1. **Support es5** - Fluture already had support for es5 by means of `require('fluture/es5')`, but this new version has it by default.
1. **Code minification** - No objections to [longer variable names](https://github.com/Avaq/Fluture/pull/30)

1. **Stack safety** - @safareli and I recently came upon the idea of making Future a flat data-structure, and interpret the structure in a stack safe way in `fork()`, much like how `ChainRec` was implemented. This allows for transforming a single Future more than `n` times, where `n` is the available space in the stack.
1. **Safe chain recursion** - Somewhat related to stack safety, the interpretation of a Future should also flatten recursive calls made via `chain()` into the same loop, allowing for infinite recursion without blowing the stack, or running out of memory.
1. **Keep the performance reasonable** - This is the reason for the Object Orientation. V8 simply does a much better job at optimizing it.

All of the above are implemented in this proof of concept. The performance has somewhat degraded compared to the previous version, but the memory footprint has somewhat improved.

----

#### TODO

See [Project 1](https://github.com/fluture-js/Fluture/projects/1)


src/core.js
```diff
@@ -0,0 +1,572 @@
...
+export class Sequence extends Core{
+
+  constructor(spawn, actions = []){
+    super();
+    this._spawn = spawn;
+    this._actions = actions;
+  }
+
+  _transform(action){
+    return new Sequence(this._spawn, this._actions.concat([action]));
```
safareli: I think using Linked List will be more efficient

safareli: and maybe in other places where you are `concat`ing on normal js array

Avaq: Good point. This is the only place I think. This array doesn't see much action though, that's all handled by the double ended queue in _fork.

src/core.js
```diff
@@ -570,7 +571,23 @@ export class Sequence extends Core{
...
+  constructor(done, value){
+    this.done = done;
+    this.value = value;
+  }
+}
+
+export const Next = x => new Iteration(false, x);
+export const Done = x => new Iteration(true, x);
+
+export const chainRec = (step, init) => (function recur(x){
```
safareli: love it ✨

src/core.js
Avaq: Any immutable list with fast `push` you know of?

safareli: don't know, I would have done something like this :d
```js
const List = daggy.taggedSum('List',{
  Cons: ['x','xs],
  Nil: []
})
```

Avaq:
Now that this is feature-complete, I'm merging it into `master`, and changed the default branch to `5.x`. [Other work](https://github.com/fluture-js/Fluture/projects/1) on the rewrite will be merged to master via individual pull requests, so reviewers can have an easier time.

PR #93, opened 2017-04-19
Update fromPromise docs to clarify that the function is called with the value

SoullessWaffle:


README.md
```diff
@@ -262,8 +262,8 @@ Furthermore; `encase2` and `encase3` are binary and ternary versions of
 Allows Promise-returning functions to be turned into Future-returning functions.
 
 Takes a function which returns a Promise, and a value, and returns a Future
-which calls the function to produce the Promise, and resolves with the Promise
-resolution value, or rejects with the Promise rejection reason.
+which calls the function with the value to produce the Promise, and resolves
+with the Promise resolution value, or rejects with the Promise rejection reason.
```
Avaq: The sentence is becoming quite long. What do you think of the following?

*Takes a function which returns a Promise, and a value, and returns a Future. When forked, the Future calls the function with the value to produce the Promise, and resolves with its resolution value, or rejects with its rejection reason*

SoullessWaffle: That sounds better, I'll update the PR.

Issue #87, opened 2017-04-19
Decide whether to flip arguments of fluent `.ap()`


Avaq:
I propose flipping the argument order of `.ap()` back to how it was pre-Fantasy Land 1.0, but we keep the argument order of `['fantasy-land/ap']()` compliant with Fantasy Land 1.0 and later. That way we tackle two problems:

1. The common complaint that the new `.ap` is awkward to use in fluent method chains.
1. The fact that Ramda and other pre-FL1 libraries cannot dispatch to `Fluture#ap`. Fluture will become compliant to *all* versions of Fantasy Land with this change.

---

The counter-argument is that this may be confusing for two reasons:

1. `Fluture#ap` will behave differently from other libraries' `ap` method. This may be a source of confusion, but it should also alleviate confusion depending on the users' expectation.
1. `Fluture#ap` will behave differently from `Fluture#fatansy-land/ap`. I'm not sure if this is confusing, since nobody uses the latter directly.

PR #96, opened 2017-05-06
Flip the argument order of fluent `ap`

Avaq:
Flipping the fluent `ap` method has some benefits, as described in #87. These should also be apparent from the changes to the README. I would like to get as many eyes as possible on this because I consider it quite a radical change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,12 +103,10 @@ For front-end applications and node <v4, please use `require('fluture/es5')`.
 [<img src="https://raw.github.com/fantasyland/fantasy-land/master/logo.png" align="right" width="82" height="82" alt="Fantasy Land" />][FL]
 [<img src="https://raw.githubusercontent.com/rpominov/static-land/master/logo/logo.png" align="right" height="82" alt="Static Land" />][6]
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -376,19 +374,20 @@ you chain more over the same structure. It's therefore recommended that you use
 traverse a large list (10000+ items).
 
 #### ap
-##### `#ap :: Future a b ~> Future a (b -> c) -> Future a c`
+##### `#ap :: Future a (b -> c) ~> Future a b -> Future a c`
```
davidchambers: Should we list type signatures for both `ap` and `fantasy-land/ap`?

Avaq: I'm not sure if `fantasy-land/ap` deserves a type signature. I believe that conceptually it's not there, and it's functions like `Z.ap` which are polymorphic.

davidchambers: My concern is that someone might see the type signature and assume that Fluture targets an old version of FL. It might be worth mentioning in the description that `fantasy-land/ap` is FL compliant.

Avaq: I've added a note about it. Do you think it's sufficient?

davidchambers: Excellent!

PR #102, opened 2017-05-12
Improve error messages related to incompatible Futures

### LINKED ISSUE #85: Improve "invalid Fluture version" error messages
Improve error messages generated when expecting a `fluture/Future@x` and getting one of the wrong version.

Avaq:
Builds on `sanctuary-type-identifiers@2.0.0`.

Closes #85

### DIFF
--- a/src/internal/const.js
+++ b/src/internal/const.js
@@ -8,3 +8,9 @@ export const FL = {
 };
 
 export const ordinal = ['first', 'second', 'third', 'fourth', 'fifth'];
+
+export const namespace = 'fluture';
+export const name = 'Future';
+export const version = 2;
+
+export const $$type = `${namespace}/${name}@${version}`;
--- a/src/internal/throw.js
+++ b/src/internal/throw.js
@@ -1,5 +1,6 @@
 import {show} from './fn';
-import {ordinal} from './const';
+import {ordinal, namespace, name, version} from './const';
+import type from 'sanctuary-type-identifiers';
 
 export const error = message => {
   throw new Error(message);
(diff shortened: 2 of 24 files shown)


src/internal/throw.js
```diff
@@ -17,3 +18,29 @@ export const invalidContext = (it, actual) => typeError(
...
+);
+
+const invalidVersion = (m, x) => (
+  `The Future was created by ${x < version ? 'an older' : 'a newer'} version of ${namespace}. `
++ 'This means that one of the sources which creates Futures is outdated. '
++ 'Update this source, or transform its created Futures to be compatible.'
++ '\n  See: https://github.com/fluture-js/Fluture#casting-futures'
+);
+
+export const invalidFuture = (it, at, m, s = '') => {
```
Avaq: I know. This function has a pretty ugly signature: `(String, String | Number, Any, String | Nil) -> Void`. It's completely unoptimizable, but I don't care, because the only time this function is called is right before the process crashes.

src/internal/const.js
```diff
@@ -8,3 +8,9 @@ export const FL = {
 };
 
 export const ordinal = ['first', 'second', 'third', 'fourth', 'fifth'];
+
+export const namespace = 'fluture';
+export const name = 'Future';
+export const version = 2;
```
davidchambers: Why `2`? Also, what is your plan for updating this value when appropriate?

Avaq: Version `2` because all Futures between 1.0.0 and 5.0.0 are conceptually version `1` (even though never explicitly specified). The version number changes when a change to the *class surface* is made. For example, all futures before 6.0.0 had an `_f`-method which Fluture expects and makes use of. Since 6.0.0 this has been renamed to `_fork`. Fluture 6.0.0 is therefore (and for more subtle reasons to do with the way Actions are implemented) incapable of consuming older Fluture Futures.

PR #109, opened 2017-05-30
Fix Static Land compliance of the TypeRep exposed by Flutures module version

Avaq:
Other than improving the code organization, this ensures that `Future.of` and `Future.reject` are present in module builds as well as UMD builds.

As a result, the following will become possible:

```diff
- import {of} from 'fluture'
- of(1)
+ import Future from 'fluture'
+ Future.of(1)
```

I find that many users expect this to work.

### DIFF
--- a/index.es.js
+++ b/index.es.js
@@ -1,12 +1,5 @@
-export {
-  Future,
-  Future as default,
-  isFuture,
-  reject,
-  of,
-  never,
-  isNever
-} from './src/core';
+export {default, default as Future} from './src/future';
+export {isFuture, reject, of, never, isNever} from './src/core';
 
 export * from './src/dispatchers';
 
--- a/src/chain-rec.js
+++ b/src/chain-rec.js
@@ -1,8 +1,7 @@
-import {Future, Core} from './core';
+import {Core} from './core';
 import {Next, Done} from './internal/iteration';
 import {Undetermined, Synchronous, Asynchronous} from './internal/timing';
 import {show, showf, noop} from './internal/fn';
-import {FL} from './internal/const';
 
 export function ChainRec(step, init){
   this._step = step;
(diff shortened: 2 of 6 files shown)


Avaq:
While reviewing my own code, I realized that for the same reason that `of` was missing from the TypeRep in module builds, so were all the other Static Land properties. In order to have the Future *module* be compliant to Static Land, I've added the appropriate dispatchers to the TypeRep.

Avaq:
Would you mind having a peek @rpominov?

index.es.js
```diff
@@ -1,12 +1,5 @@
-export {
-  Future,
-  Future as default,
-  isFuture,
-  reject,
-  of,
-  never,
-  isNever
-} from './src/core';
+export {default, default as Future} from './src/future';
```
rpominov: Not 100% sure that ES specification allows to do `export {default} from '...'`. It maybe the case that it works with current tools by incident, and may not work in the future.

Avaq: Reading through [the spec](https://tc39.github.io/ecma262/#prod-ExportDeclaration), it seems to be alright. One concerning rule is the following:

> For each *IdentifierName* `n` in ReferencedBindings of *ExportClause*: It is a Syntax Error if StringValue of `n` is a *ReservedWord*

In the [spec for *ReservedWord*](https://tc39.github.io/ecma262/#prod-ReservedWord) we see that `default` *is* actually a reserved word. This would mean that `export {Future as default} from '...'` also shouldn't work.

rpominov: Right. Maybe you should do `import Future from './src/future'; export default Future;` to be safe.

Avaq: It means I have to do it like this:

```js
import _Future from './src/future';

export default _Future;
export const Future = _Future;
```

rpominov: Yeah, ugly. But I would probably do it like this though.

Avaq: I've made the same mistake in other places of the code. I'll merge this PR as is, and create a separate issue to discuss the matter.

PR #113, opened 2017-06-05
Reorganize the API documentation

Avaq:
The goals of this change:

* Not to break any links
* To move logo and sponsorship credits to a more prominent place
* To improve the ordering of documentation sections by moving related information closer together
* To make the table of contents more useful and less daunting

See [the rendered HTML](https://github.com/fluture-js/Fluture/blob/avaq/api-docs/README.md).

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,7 @@ Some of the features provided by Fluture include:
 
 For more information:
 
+* [API documentation](#documentation)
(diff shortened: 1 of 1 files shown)


davidchambers:
Whoa, `<details>` is really nice! I didn't know this behaviour could be achieved with HTML alone!

PR #115, opened 2017-06-09
Fix undefined action in interpretor

stevenyap:
I faced an edge case where action is undefined.
I could not replicate it successfully but found a workaround
to check if action is defined before calling rejected on action.

This error only comes up when I am running Jest on a code where
Future.parallel is used and there are two rejected futures.

Error:
fluture/index.js:155
settle(action.rejected(x));
             ^
TypeError: Cannot read property 'rejected' of undefined


stevenyap:
I suspect it could be due to some concurrent processes running
resulting in a missing action subsequent but could not get my head
around to find the cause.

Anyway, I figures that it is always good to check for undefined so here's the PR.

Avaq:
Hi Steven. Thank you for reporting the issue, and your effort to resolve it! Instead of adding in null-checks, let's get to the bottom of this. I'd like to ask you to revert the changes made thus far (including the `yarn.lock` file), and instead focus your efforts towards adding a failing unit test which reproduces your edge-case. I'll add a comment to each of your changed lines explaining why I'd like it to be reverted.

Feel free to install additional libraries if they are needed to reproduce the bug. See [this bug test](https://github.com/fluture-js/Fluture/blob/5b62f7b7ee2ead0d81e1133de026843b080956bc/test/2.sequence.test.js#L394-L404) for an example of what I mean. Once we have a failing test, we can start to reason about what's really going wrong.

src/internal/interpretor.js
```diff
@@ -29,7 +29,7 @@ export default function interpretor(rej, res){
   function settle(m){
     settled = true;
     future = m;
-    if(future._spawn){
+    if(future && future._spawn){
```
Avaq: Why this line should be reverted: There is no need to test for the presence of `future` here. We have *just* assigned the function argument `m` to `future`. Since it is our intention to call this function with an instance of Future always, `future` should always be an there. Besides, if it's not here then our interpretor will fail in the next tick because it's `settle`s job to prepare an instance of `future` to be used in the next tick.

src/internal/interpretor.js
```diff
@@ -40,14 +40,14 @@ export default function interpretor(rej, res){
   //It will tell the current action that the future rejected, and it will
   //settle the current tick with the action's answer to that.
   function rejected(x){
-    settle(action.rejected(x));
+    settle(action && action.rejected(x));
```
Avaq: Why this line should be reverted: This line is now buggy. If action would be falsy, we would be passing a falsy value into `settle` which will lead to more breakage. We *could* "fix" it like this: `action && settle(action.rejected(x))`, but that will cause the Future to hang when the action is missing.

More importantly there should be no code paths allowing for `action` to be anything else than an instance of `Action`. I suspect from your bug report that there actually *is* a code path like that hiding somewhere, which means we should find it and fix it, rather than implementing a null check.

src/internal/interpretor.js
```diff
@@ -40,14 +40,14 @@ export default function interpretor(rej, res){
...
   }
 
   //This function serves as a resolution handler for our current future.
   //It will tell the current action that the future resolved, and it will
   //settle the current tick with the action's answer to that.
   function resolved(x){
-    settle(action.resolved(x));
+    settle(action && action.resolved(x));
```
Avaq: This line should be reverted for the same reason the `rejected` one should.

Avaq:
Let's also make sure we ruled out a few basic problems:

* Are you using the latest version of Fluture (6.2.6)?
* Does the bug occur if you install your dependencies with npm?
* Which version of Node are you using?

Avaq:
> This error only comes up when I am running Jest on a code where Future.parallel is used and there are two rejected futures.

In theory, concurrently running Futures have no influence over each others interpreters, so you *should* be able to locate the faulting Future and run it on its own (without anything else parallel to it), and it should still produce the same result.

stevenyap:
Thanks for explaining the revert clearly.
I would have to dive into the code base of Fluture in order to replicate the bug successfully.
As this would take more time, I will close this for now and revert once I have managed to replicate the bug using the clue you have given:

> EDIT: The only way they influence other interpretation loops is by calling the cancel. This may have been happening since you mentioned some of the Futures given to parallel rejected.

PR #122, opened 2017-06-30
Improve the documentation regarding type signatures

Avaq:
* Add a section explaining the Hindley-Milner extensions.
* Add missing concrete types to the list of types.
* Use Haskell code blocks instead of headers to document function signatures.
* Hide extended function signature information in `<details>` blocks.
* Align type signature of function variations so that it's easier to recognize the similarities.

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,11 +185,53 @@ sponsoring the project.
 
 ### Type signatures
 
-[Hindley-Milner][Guide:HM] type signatures are used to document functions, with
-one addition: In order to document *methods* as well as static functions, we
-prefix every signature with `.` or `#`. This indicates whether the function
-lives on the object (`.function`), or its prototype (`#method`). We use `~>` to
-separate the implicit `this` argument from the other, explicit, arguments.
+[Hindley-Milner][Guide:HM] type signatures are used to document functions. You
+might encounter some additional syntax that we use to describe JavaScript
+specific stuff, like [methods](#squiggly-arrows) or functions that take
+[multiple arguments at once](#brackets).
+
+#### Squiggly Arrows
+
+In order to document *methods*, we use the squiggly arrow (`~>`). This separates
+the implicit `this` argument from the other, explicit, arguments. For example,
+the following line signifies a *function*, because it doesn't have a squiggly
+arrow in its signature:
+
+```hs
+map :: (b -> c) -> Future a b -> Future a c
+```
+
+Whereas the next example is a *method*. It needs a `this` as indicated by the
+use of the squiggly arrow.
+
+```hs
+Future.prototype.map :: Future a b ~> (b -> c) -> Future a c
+```
+
+#### Brackets
+
+Most functions exposed by Fluture are curried. This is reflected in their type
+signatures by using an arrow at each step where partial application is possible.
+For example, the following line signifies a *curried* function, because it has
+an arrow after each function argument:
+
+```hs
+add :: Number -> Number -> Number
+```
+
+We could have chosen to write the above line with "groups of one argument", but
+we usually leave the grouping brackets out for brevity:
+
+```hs
+add :: (Number) -> (Number) -> Number
+```
+
+In order to document functions and methods that are *not* curried, we use
+grouping to show which arguments have to be provided at the same time:
+
+```hs
+add :: (Number, Number) -> Number
+```
 
 #### Types
 
@@ -199,9 +241,11 @@ This is reference of the types used throughout the documentation:
 - **Future** - Instances of Future provided by [compatible versions](#casting-futures) of Fluture.
 - **ConcurrentFuture** - [Concurrified][concurrify] Futures ([`Future.Par`](#concurrentfuture)).
 - **Promise** - Values which conform to the [Promises/A+ specification][7].
-- **Nodeback a b** - A Node-style callback; A function of signature `(a | Nil, b) -> ()`.
+- **Nodeback a b** - A Node-style callback; A function of signature `(a | Nil, b) -> x`.
+- **Pair a b** - An array with exactly two elements: `[a, b]`.
 - **Iterator** - Objects with `next`-methods which conform to the [Iterator protocol][3].
 - **Cancel** - The nullary [cancellation](#future) functions returned from computations.
+- **Catchable e f** - A function `f` which may throw an exception `e`.
 
 #### Type classes
 
@@ -300,7 +344,14 @@ compatible.both(Future.of('world')).value(console.log);
 ### Creating Futures
 
 #### Future
-##### `Future :: ((a -> (), b -> ()) -> Cancel) -> Future a b`
+
+<details><summary><code>Future :: ((a -> x, b -> x) -> Cancel) -> Future a b</code></summary>
+
+```hs
+Future :: ((a -> x, b -> x) -> Cancel) -> Future a b
+```
+
+</details>
 
 Creates a Future with the given computation. A computation is a function which
 takes two callbacks. Both are continuations for the computation. The first is
@@ -322,7 +373,15 @@ cancellation logic. This function is executed when the Future is cancelled
 after it's [forked](#fork).
 
 #### of
-##### `.of :: a -> Future _ a`
+
+<details><summary><code>of :: b -> Future a b</code></summary>
+
+```hs
+of        :: b -> Future a b
+Future.of :: b -> Future a b
+```
+
+</details>
 
 Creates a Future which immediately resolves with the given value. This function
 is compliant with the [Fantasy Land Applicative specification][FL:applicative].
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -185,21 +185,66 @@ sponsoring the project.
...
-lives on the object (`.function`), or its prototype (`#method`). We use `~>` to
-separate the implicit `this` argument from the other, explicit, arguments.
+[Hindley-Milner][Guide:HM] type signatures are used to document functions. You
+might encounter some additional syntax that we use to describe JavaScript
+specific stuff, like [methods](#squiggly-arrows) or functions that take
+[multiple arguments at once](#brackets).
+
+#### Squiggly Arrows
+
+In order to document *methods* as well as static functions, use the squiggly
```
davidchambers: This would be clearer without _as well as static functions_, I think. Also, should it be _we use_ rather than _use_?

README.md
```diff
@@ -185,21 +185,66 @@ sponsoring the project.
...
+```hs
+map :: Future a b ~> (b -> c) -> Future a c
+```
+
+#### Brackets
+
+Most functions exposed by Fluture are curried. This is reflected in their type
+signatures by using an arrow at each step where partial application is possible.
+For example, the following line signifies a *curried* function, because it has
+an arrow in between every function argument:
```
davidchambers: I think _after each_ would be clearer than _in between every_.

README.md
```diff
@@ -185,21 +185,66 @@ sponsoring the project.
...
+```hs
+add :: (Number, Number) -> Number
+```
 
 #### Types
 
 You'll find that some signatures refer to concrete types, such as `Future`.
 This is reference of the types used throughout the documentation:
 
+- **()** - Nothing, also known as "unit". Corresponds to `undefined`.
```
davidchambers: I would avoid using the word _nothing_ as it could cause readers to think of the Maybe type.

Why did you decide to use `()` rather than `Undefined`? In addition to being cryptic to those unfamiliar with Haskell, it introduces ambiguity: `() -> Interger` could either be a nullary function which returns an integer or a unary function which takes `undefined :: ()` and returns an integer.

README.md
```diff
@@ -185,21 +185,66 @@ sponsoring the project.
...
 - **Future** - Instances of Future provided by [compatible versions](#casting-futures) of Fluture.
 - **ConcurrentFuture** - [Concurrified][concurrify] Futures ([`Future.Par`](#concurrentfuture)).
 - **Promise** - Values which conform to the [Promises/A+ specification][7].
 - **Nodeback a b** - A Node-style callback; A function of signature `(a | Nil, b) -> ()`.
+- **Tuple a b** - An array with exactly two elements: `[a, b]`.
```
davidchambers: Consider using `Pair a b` instead to match Sanctuary (see [`S.pairs`][1], for example).


[1]: https://sanctuary.js.org/#pairs

README.md
```diff
@@ -300,7 +345,14 @@ compatible.both(Future.of('world')).value(console.log);
 ### Creating Futures
 
 #### Future
-##### `Future :: ((a -> (), b -> ()) -> Cancel) -> Future a b`
```
davidchambers: It's worth mentioning that this change will partially break links. I don't think this is a problem, though.

README.md
```diff
@@ -322,7 +374,14 @@ cancellation logic. This function is executed when the Future is cancelled
 after it's [forked](#fork).
 
 #### of
-##### `.of :: a -> Future _ a`
+
+<details><summary><code>of :: b -> Future a b</code></summary>
+
+```hs
+of :: b -> Future a b
```
davidchambers: `b -> Future a b` reads better than `a -> Future _ a`. :thumbsup:

README.md
```diff
@@ -413,8 +500,15 @@ Future.do(function*(){
...
 
 #### try
-##### `.try :: (() -> !a | b) -> Future a b`
-##### `.attempt :: (() -> !a | b) -> Future a b`
+
+<details><summary><code>try :: (() -> !a | b) -> Future a b</code></summary>
+
+```hs
+try     :: (() -> !a | b) -> Future a b
+attempt :: (() -> !a | b) -> Future a b
```
davidchambers: Is the meaning of `!a` explained somewhere? Is the meaning of `a | b` explained somewhere?

README.md
```diff
@@ -431,7 +525,14 @@ Future.try(() => data.foo.bar.baz)
 This function has an alias `attempt`, for environments in which `try` is a reserved word.
 
 #### tryP
-##### `.tryP :: (() -> Promise e r) -> Future e r`
+
+<details><summary><code>tryP :: (() -> Promise e r) -> Future e r</code></summary>
+
+```hs
+tryP :: (() -> Promise e r) -> Future e r
```
davidchambers: `()` is ambiguous here, due to `undefined :: ()`. I suggest using `Undefined` to represent the type of `undefined` (as sanctuary-def does).

README.md
```diff
@@ -465,9 +573,16 @@ Future.node(done => {
...
-##### `.encase :: (a -> !e | r) -> a -> Future e r`
-##### `.encase2 :: (a, b -> !e | r) -> a -> b -> Future e r`
-##### `.encase3 :: (a, b, c -> !e | r) -> a -> b -> c -> Future e r`
+
+<details><summary><code>encase  :: (Catchable e (a -> r)) -> a -> Future e r</code></summary>
+
+```hs
+encase  :: (Catchable e ((a      ) -> r)) -> a ->           Future e r
+encase2 :: (Catchable e ((a, b   ) -> r)) -> a -> b ->      Future e r
+encase3 :: (Catchable e ((a, b, c) -> r)) -> a -> b -> c -> Future e r
```
davidchambers: :heart:

README.md
```diff
@@ -556,8 +692,15 @@ function directly. Instead it's recommended to use [`chain(rec)`](#chain).
...
 
 #### map
-##### `#map :: Future a b ~> (b -> c) -> Future a c`
-##### `.map :: Functor m => (a -> b) -> m a -> m b`
+
+<details><summary><code>map :: Future e a ~> (a -> b) -> Future e b</code></summary>
+
+```hs
+map :: Future e a ~> (a -> b)        -> Future e b
+map :: Functor m =>  (a -> b) -> m a -> m        b
```
davidchambers: Shall we align `=>` with `~>`?

README.md
```diff
@@ -1016,10 +1285,17 @@ const closeConnection = conn => Future((rej, res) => {
...
-##### `.finally :: Future a c -> Future a b -> Future a b`
-##### `.lastly :: Future a c -> Future a b -> Future a b`
+
+<details><summary><code>finally :: Future a b ~> Future a c -> Future a b</code></summary>
+
+```hs
+finally :: Future a b ~> Future a c               -> Future a b
+lastly  :: Future a b ~> Future a c               -> Future a b
+finally ::               Future a c -> Future a b -> Future a b
+lastly  ::               Future a c -> Future a b -> Future a b
```
davidchambers: This gives the impression that `finally` and `lastly` have two types each. The `#`/`.` served a useful role, I think.

README.md
Avaq: Good catch. This should be changed to `Catchable a (() -> b)`, as I did with `encase`.

README.md
Avaq: The `#`/`.` are redundant, as the `~>` already indicates where a function lives. I think I should remove aliases from these though, as I believe they may add to the confusion.

README.md
Avaq: Yeah, those lines were never intended to be used as links, and the documentation never linked to them. I'm okay with breaking the deep links out there that people may have generated, I don't expect there to be many.

README.md
Avaq: Alternatively, I can go with the Sanctuary approach of naming the function exactly after where it lives, so `prototype.finally`, for example. Can you look at other functions (like `fork`) and see if those are also confusing in their current form?

README.md
Avaq: Actually I can just replace instances of `-> ()` by `-> x` and I won't need a special `Undefined` type. The return value of these functions can be anything, but is ignored.

davidchambers: :thumbsup:

README.md
davidchambers: I find it confusing for an expression to have different meanings depending on the type signatures associated with it. I would find `Future.finally`/`Future.prototype.finally` much clearer than `finally`/`finally`.

Avaq:
What do you think of the usage of `<details>` in this PR?

After putting signatures in code-blocks, I realized they take up a lot of space, but especially that they draw too much attention. The solution was to use `<details>`-blocks, but I'm also aware that *I'll solve it with `<details>`-blocks* is a thought I've had just a little too often lately.

davidchambers:
> What do you think of the usage of `<details>` in this PR?

I'm positive overall, but I have two concerns:

  - in some cases the revealed text is identical to the visible text; and
  - on GitHub the collapsed signatures appear too close to the subsequent paragraphs.

README.md
Avaq: `Future.finally` would be misleading in the context of the ES6 module, as the named import `Future` has no property `finally`. Instead, `finally` is a named import that lives alongside `Future`.

davidchambers: Oh, I see! In that case I suggest `finally`/`Future.prototype.finally`.

Avaq:
Alright, I made the change to document where functions live explicitly. Note that `of`, `reject`, `ap`, `map`, `bimap` and `chain` are special because they exist in three places:

* As a module export
* On the `Future` typeRep to support Static Land
* On the Future prototype to support "fluent" syntax

PR #124, opened 2017-07-06
Prevent Future.parallel from cancelling settled computations

### LINKED ISSUE #123: Undefined action in interpretor
Related: #115 
@stevenyap ran into this issue a while ago, but I was not able to reproduce it. I *just* experienced the same issue in one of my own projects, so I will isolate it and fix it.

```text
.../node_modules/fluture/index.js:162
    settle(action.resolved(x));
                 ^

TypeError: Cannot read property 'resolved' of undefined
    at Timeout.resolved (.../node_modules/fluture/index.js:162:18)
    at Timeout.ontimeout [as _onTimeout] (timers.js:341:34)
    at tryOnTimeout (timers.js:232:11)
    at Timer.listOnTimeout (timers.js:202:5)
```

Avaq:
This was a terribly difficult bug to track down. You may be interested to see this @stevenyap:

* `Future.parallel` sends a cancel signal to futures that reject
* Some `_fork` methods cannot deal with that. In particular: `Future.hook` does not expect to be cancelled right after being rejected, and this lead to an unwanted resolution signal being sent.
* The resolution signal would cause problems in the sequence interpreter. In particular: The interpreter would already be "done" and cleaned up its state when receiving the unexpected resolution signal. This in turn would cause it to try and re-resolve the current action, which finally left us with `action.resolve` throwing an error.

Fixes #123

### DIFF
--- a/src/parallel.js
+++ b/src/parallel.js
@@ -1,6 +1,6 @@
 import {Core, Resolved, isFuture} from './core';
 import {invalidArgument, invalidFuture} from './internal/throw';
-import {show, mapArray, partial1} from './internal/fn';
+import {noop, show, mapArray, partial1} from './internal/fn';
 import {isUnsigned, isArray} from './internal/is';
 
 const check$parallel = (m, i) => isFuture(m) ? m : invalidFuture(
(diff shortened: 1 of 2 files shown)


davidchambers:
Nice pull request description, Aldwin! Making such notes contemporaneously is a good idea. :)

PR #128, opened 2017-07-13
Improve the performance of composing computations

Avaq:
Instead of using an array, and forcing it to be "immutable" by using `.concat([item])`, I've defined a very simple immutable linked list and use that to build up the list of operations to perform during interpretation.

As an example, I took the [*stack safety* example](https://github.com/fluture-js/Fluture#stack-safety) from Fluture's own README:

```js
//test.js
const Future = require('.');

const add1 = x => x + 1;
let m = Future.of(1);

for(let i = 0; i < 100000; i++){
  m = m.map(add1);
}

m.fork(console.error, console.log);
```

**Before**

```sh
$ time node test.js 
100001

real    0m30.145s
user    0m23.878s
sys     0m8.083s
```

**After**

```sh
$ time node test.js 
100001

real    0m0.153s
user    0m0.148s
sys     0m0.019s
```

For this type of script, the performance improves significantly. In practice, this improvement will probably be encountered in scripts using this pattern I've seen here and there:

```js
reduce((m, x) => m.chain(acc => reducer(acc, x)), Future.of(initial), xs)
```

The general performance as measured by the Doxbee testing suite has improved by about 11%.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -3,6 +3,7 @@ import {isFunction} from './internal/is';
 import {error, typeError, invalidArgument, invalidContext, invalidFuture} from './internal/throw';
 import {$$type} from './internal/const';
 import interpreter from './internal/interpreter';
+import {empty as emptyList, cons} from './internal/list';
 import type from 'sanctuary-type-identifiers';
 
 const throwRejection = x => error(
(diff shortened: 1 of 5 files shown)


src/internal/list.js
```diff
@@ -0,0 +1,22 @@
+function List(head, tail){
+  this.head = head;
+  this.tail = tail;
+  this.size = tail.size + 1;
+}
```
Avaq: @safareli, you once suggested using a List over an Array. You may be interesting in seeing it. :)

src/internal/const.js
```diff
@@ -11,6 +11,6 @@ export const ordinal = ['first', 'second', 'third', 'fourth', 'fifth'];
 
 export const namespace = 'fluture';
 export const name = 'Future';
-export const version = 2;
+export const version = 3;
```
Avaq: @davidchambers Note how I'm incrementing the version of the Fluture `@@type`. This is because in changing the internal structure of the `Sequence` instances, I've broken backwards compatibility. `interpret` expects `_actions` to be a `List` now, and not an `Array`.

Despite this *technically* being a "breaking change", I will not release it as a major version, but use a minor version instead. Detecting incompatible versions is really just there to improve error messages.

src/internal/list.js
```diff
@@ -0,0 +1,22 @@
...
+
+List.prototype.isEmpty = false;
+
+List.prototype.add = function List$add(x){
```
davidchambers: Why `add` rather than `cons`?

src/internal/list.js
```diff
@@ -0,0 +1,22 @@
...
+function Empty(){}
+
+Empty.prototype = Object.create(List.prototype);
+
+Empty.prototype.isEmpty = true;
+Empty.prototype.head = null;
+Empty.prototype.tail = null;
+Empty.prototype.size = 0;
+
+export default new Empty;
```
davidchambers: I find this export confusing. I would not expect `require('./internal/list')` to give me this value.

src/core.js
```diff
@@ -563,5 +564,12 @@ Sequence.prototype._finally = function Sequence$finally(other){
 Sequence.prototype._fork = interpreter;
 
 Sequence.prototype.toString = function Sequence$toString(){
-  return `${this._spawn.toString()}${this._actions.map(x => `.${x.toString()}`).join('')}`;
+  let str = this._spawn.toString(), tail = this._actions;
+
+  while(!tail.isEmpty){
+    str = `${str}.${tail.head.toString()}`;
```
davidchambers: Should we use `Z.toString` here?

Avaq: We can safely avoid unnecessary dispatching, because we *know* that every item in `this._actions` has a custom `toString` function defined which we want to use. If not, that would be a bug in Fluture.

src/internal/list.js
Avaq: Because I have no idea about conventions. ;)
:zap:

src/internal/list.js
Avaq: It's because I'm implementing the bare minimum, so to construct a list one does `list.add(1).add(2).add(3)`. No need to export the internals, or a specialized constructor.

davidchambers: I suggest renaming the file __empty.js__, in that case.

safareli: You don't need Empty constructor, can just have special object created from List.prototype with desired props.

src/internal/list.js
safareli: nice!

src/internal/list.js
```diff
@@ -0,0 +1,18 @@
...
+List.prototype.cons = function List$cons(x){
+  return new List(x, this);
+};
+
+const empty = Object.create(List.prototype);
+
+empty.isEmpty = true;
+empty.size = 0;
+
+export default empty;
```
Avaq: @davidchambers I'm still quite fond of exporting `empty` from `list`. Essentially we're exporting `[]` and `:` (`cons`), and those are all the necessary building blocks for creating lists. Not only do I feel like this is elegant in a way, it's also quite convenient because I don't have to do anything before using it the way I want to in `core.js`.

src/internal/list.js
```diff
@@ -0,0 +1,2 @@
+export const empty = ({isEmpty: true, size: 0, head: null, tail: null});
+export const cons = (head, tail) => ({isEmpty: false, size: tail.size + 1, head, tail});
```
Avaq: Surprisingly, defining the list type like this has zero impact on performance. What do you think @davidchambers and @safareli? I think it addresses both of your points of feedback. :)

davidchambers: Beautiful! :unicorn:

The last thing to consider is `s/empty/nil/` to avoid confusion with `Z.empty`.

The parens around the object literal assigned to `empty` are not required (although you may have included them for aesthetic reasons).

Avaq: I did include them for aesthetic reasons. :)
I'll keep it as empty, as I consider it more descriptive within the context of a list. I renamed it `emptyList` during import to `core.js`, to preserve context.

safareli: nice

PR #129, opened 2017-07-16
Add a TypeScript definition file

tusharmath:
@Avaq 
I have created the PR but could use some help since there are a lot utility functions to give typings for.

### TODO

- [x] `after()`
- [x] `and()`
- [x] `ap()`
- [x] `attempt()`
- [x] `bimap()`
- [x] `both()`
- [x] `cache()`
- [x] `chain()`
- [x] `done()`
- [x] `encase()`
- [x] `encase2()`
- [x] `encase3()`
- [x] `encaseN()`
- [x] `encaseN2()`
- [x] `encaseN3()`
- [x] `encaseP()`
- [x] `encaseP2()`
- [x] `encaseP3()`
- [x] `extractLeft()`
- [x] `extractRight()`
- [x] `fold()`
- [x] `fork()`
- [x] `go()`
- [x] `hook()`
- [x] `isFuture()`
- [x] `isNever()`
- [x] `lastly()`
- [x] `map()`
- [x] `mapRej()`
- [x] `never()`
- [x] `node()`
- [x] `of()`
- [x] `or()`
- [x] `par()`
- [x] `parallel()`
- [x] `promise()`
- [x] `race()`
- [x] `reject()`
- [x] `rejectAfter()`
- [x] `seq()`
- [x] `swap()`
- [x] `tryP()`
- [x] `value()`
- [x] `Future()`

-----

- [x] Implement typings for all methods on the Future interface
- [x] Add a line of documentation to all functions and methods, with links to online docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,8 @@ getPackageName('package.json')
 
 ### In ES6 or newer environments
 
-The `package.json` sets a `module`-field for build-tools like [Rollup][].
+The `package.json` sets a `module`-field for build-tools like [Rollup][]. The
+module version also has TypeScript definitions.
 
 ```js
 import {readFile} from 'fs';
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,320 @@
+declare module 'fluture' {
+
+  interface RejectFunction<L> {
+    (error: L): void
+  }
+
+  interface ResolveFunction<R> {
+    (value: R): void
+  }
+
+  interface Cancel {
+    (): void
+  }
+
+  interface Nodeback<E, R> {
+    (err: E | null, value?: R): void
+  }
+
+  interface Next<T> {
+    done: false
+    value: T
+  }
+
+  interface Done<T> {
+    done: true
+    value: T
+  }
+
+  interface Iterator<N, D> {
+    next(value?: N): Next<N> | Done<D>
+  }
+
+  interface Generator<Y, R> {
+    (): Iterator<Y, R>
+  }
+
+  /** The function is waiting for two more arguments. */
+  interface AwaitingTwo<A, B, R> {
+    (a: A, b: B): R
+    (a: A): (b: B) => R
+  }
+
+  /** The function is waiting for three more arguments. */
+  interface AwaitingThree<A, B, C, R> {
+    (a: A, b: B, c: C): R
+    (a: A, b: B): (c: C) => R
+    (a: A): AwaitingTwo<B, C, R>
+  }
+
+  export interface ConcurrentFuture<L, R> {
+    sequential: Future<L, R>
+  }
+
+  export interface Future<L, R> {
+
+    /** Logical and for Futures. See https://github.com/fluture-js/Fluture#and */
+    and<RB>(right: Future<L, RB>): Future<L, RB>
+
+    /** Apply the function in this Future to the value in another. See https://github.com/fluture-js/Fluture#ap */
+    ap<A, B>(this: Future<L, (value: A) => B>, right: Future<L, A>): Future<L, B>
+
+    /** Map over both branches of the Future at once. See https://github.com/fluture-js/Fluture#bimap */
+    bimap<LB, RB>(lmapper: (reason: L) => LB, rmapper: (value: R) => RB): Future<LB, RB>
+
+    /** Wait for this Future and the given one in parallel. See https://github.com/fluture-js/Fluture#both */
+    both<RB>(right: Future<L, RB>): Future<L, [R, RB]>
+
+    /** Use the resolution value in this Future to create the next Future. See https://github.com/fluture-js/Fluture#chain */
+    chain<RB>(mapper: (value: R) => Future<L, RB>): Future<L, RB>
+
+    /** Use the rejection reason in this Future to create the next Future. See https://github.com/fluture-js/Fluture#chainrej */
+    chainRej<LB>(mapper: (reason: L) => Future<LB, R>): Future<LB, R>
+
+    /** The Future constructor */
+    constructor: Fluture
+
+    /** Fork the Future into a Node-style callback. See https://github.com/fluture-js/Fluture#done */
+    done(callback: Nodeback<L, R>): Cancel
+
+    /** Set up a cleanup Future to run after this one is done. See https://github.com/fluture-js/Fluture#finally */
+    finally(cleanup: Future<L, any>): Future<L, R>
+
+    /** Fold both branches into the resolution branch. See https://github.com/fluture-js/Fluture#fold */
+    fold<RB>(lmapper: (reason: L) => RB, rmapper: (value: R) => RB): Future<never, RB>
+
+    /** Fork the Future into the two given continuations. See https://github.com/fluture-js/Fluture#fork */
+    fork(reject: RejectFunction<L>, resolve: ResolveFunction<R>): Cancel
+
+    /** Set up a cleanup Future to run after this one is done. See https://github.com/fluture-js/Fluture#finally */
+    lastly(cleanup: Future<L, any>): Future<L, R>
+
+    /** Map over the resolution value in this Future. See https://github.com/fluture-js/Fluture#map */
+    map<RB>(mapper: (value: R) => RB): Future<L, RB>
+
+    /** Map over the rejection reason in this Future. See https://github.com/fluture-js/Fluture#maprej */
+    mapRej<LB>(mapper: (reason: L) => LB): Future<LB, R>
+
+    /** Logical or for Futures. See https://github.com/fluture-js/Fluture#or */
+    or(right: Future<L, R>): Future<L, R>
+
+    /** Fork the Future into a Promise. See https://github.com/fluture-js/Fluture#promise */
+    promise(): Promise<R>
+
+    /** Race this Future against another one. See https://github.com/fluture-js/Fluture#race */
+    race(right: Future<L, R>): Future<L, R>
+
+    /** Swap the rejection reason and the resolotion value. See https://github.com/fluture-js/Fluture#swap */
+    swap(): Future<R, L>
+
+    /** Fork this Future into the given continuation. See https://github.com/fluture-js/Fluture#value */
+    value(this: Future<never, R>, resolve: ResolveFunction<R>): Cancel
+
+  }
+
+  /** Creates a Future which resolves after the given duration with the given value. See https://github.com/fluture-js/Fluture#after */
+  export function after<R>(duration: number, value: R): Future<never, R>
+  export function after<R>(duration: number): (value: R) => Future<never, R>
+
+  /** Logical and for Futures. See https://github.com/fluture-js/Fluture#and */
+  export function and<L, R>(left: Future<L, any>, right: Future<L, R>): Future<L, R>
+  export function and<L, R>(left: Future<L, any>): (right: Future<L, R>) => Future<L, R>
+
+  /** Apply the function in the left Future to the value in the right Future. See https://github.com/fluture-js/Fluture#ap */
+  export function ap<L, RA, RB>(apply: Future<L, (value: RA) => RB>, value: Future<L, RA>): Future<L, RB>
+  export function ap<L, RA, RB>(apply: Future<L, (value: RA) => RB>): (value: Future<L, RA>) => Future<L, RB>
+
+  /** Create a Future which resolves with the return value of the given function, or rejects with the error it throws. See https://github.com/fluture-js/Fluture#try */
+  export function attempt<L, R>(fn: () => R): Future<L, R>
+
+  /** Map over both branched of the given Bifunctor at once. See https://github.com/fluture-js/Fluture#bimap */
+  export function bimap<LA, LB, RA, RB>(lmapper: (reason: LA) => LB, rmapper: (value: RA) => RB, source: Future<LA, RA>): Future<LB, RB>
+  export function bimap<LA, LB, RA, RB>(lmapper: (reason: LA) => LB, rmapper: (value: RA) => RB): (source: Future<LA, RA>) => Future<LB, RB>
+  export function bimap<LA, LB, RA, RB>(lmapper: (reason: LA) => LB): (rmapper: (value: RA) => RB, source: Future<LA, RA>) => Future<LB, RB>
+  export function bimap<LA, LB, RA, RB>(lmapper: (reason: LA) => LB): (rmapper: (value: RA) => RB) => (source: Future<LA, RA>) => Future<LB, RB>
+
+  /** Wait for both Futures to resolve in parallel. See https://github.com/fluture-js/Fluture#both */
+  export function both<L, A, B>(left: Future<L, A>, right: Future<L, B>): Future<L, [A, B]>
+  export function both<L, A, B>(left: Future<L, A>): (right: Future<L, B>) => Future<L, [A, B]>
+
+  /** Cache the outcome of the given Future. See https://github.com/fluture-js/Fluture#cache */
+  export function cache<L, R>(source: Future<L, R>): Future<L, R>
+
+  /** Create a Future using the resolution value of the given Future. See https://github.com/fluture-js/Fluture#chain */
+  export function chain<L, RA, RB>(mapper: (value: RA) => Future<L, RB>, source: Future<L, RA>): Future<L, RB>
+  export function chain<L, RA, RB>(mapper: (value: RA) => Future<L, RB>): (source: Future<L, RA>) => Future<L, RB>
+
+  /** Create a Future using the rejection reason of the given Future. See https://github.com/fluture-js/Fluture#chain */
+  export function chainRej<LA, LB, R>(mapper: (reason: LA) => Future<LA, R>, source: Future<LA, R>): Future<LB, R>
+  export function chainRej<LA, LB, R>(mapper: (reason: LA) => Future<LA, R>): (source: Future<LA, R>) => Future<LB, R>
+
+  /** Fork the given Future into a Node-style callback. See https://github.com/fluture-js/Fluture#done */
+  export function done<L, R>(callback: Nodeback<L, R>, source: Future<L, R>): Cancel
+  export function done<L, R>(callback: Nodeback<L, R>): (source: Future<L, R>) => Cancel
+
+  /** Encase the given function such that it returns a Future of its return value. See https://github.com/fluture-js/Fluture#encase */
+  export function encase<L, R, A>(fn: (a: A) => R, a: A): Future<L, R>
+  export function encase<L, R, A>(fn: (a: A) => R): (a: A) => Future<L, R>
+
+  /** Encase the given function such that it returns a Future of its return value. See https://github.com/fluture-js/Fluture#encase */
+  export function encase2<L, R, A, B>(fn: (a: A, b: B) => R, a: A, b: B): Future<L, R>
+  export function encase2<L, R, A, B>(fn: (a: A, b: B) => R, a: A): (b: B) => Future<L, R>
+  export function encase2<L, R, A, B>(fn: (a: A, b: B) => R): AwaitingTwo<A, B, Future<L, R>>
+
+  /** Encase the given function such that it returns a Future of its return value. See https://github.com/fluture-js/Fluture#encase */
+  export function encase3<L, R, A, B, C>(fn: (a: A, b: B, c: C) => R, a: A, b: B, c: C): Future<L, R>
+  export function encase3<L, R, A, B, C>(fn: (a: A, b: B, c: C) => R, a: A, b: B): (c: C) => Future<L, R>
+  export function encase3<L, R, A, B, C>(fn: (a: A, b: B, c: C) => R, a: A): AwaitingTwo<B, C, Future<L, R>>
+  export function encase3<L, R, A, B, C>(fn: (a: A, b: B, c: C) => R): AwaitingThree<A, B, C, Future<L, R>>
+
+  /** Encase the given Node-style function such that it returns a Future of its result. See https://github.com/fluture-js/Fluture#encasen */
+  export function encaseN<L, R, A>(fn: (a: A, callback: Nodeback<L, R>) => void, a: A): Future<L, R>
+  export function encaseN<L, R, A>(fn: (a: A, callback: Nodeback<L, R>) => void): (a: A) => Future<L, R>
+
+  /** Encase the given Node-style function such that it returns a Future of its result. See https://github.com/fluture-js/Fluture#encasen */
+  export function encaseN2<L, R, A, B>(fn: (a: A, b: B, callback: Nodeback<L, R>) => void, a: A, b: B): Future<L, R>
+  export function encaseN2<L, R, A, B>(fn: (a: A, b: B, callback: Nodeback<L, R>) => void, a: A): (b: B) => Future<L, R>
+  export function encaseN2<L, R, A, B>(fn: (a: A, b: B, callback: Nodeback<L, R>) => void): AwaitingTwo<A, B, Future<L, R>>
+
+  /** Encase the given Node-style function such that it returns a Future of its result. See https://github.com/fluture-js/Fluture#encasen */
+  export function encaseN3<L, R, A, B, C>(fn: (a: A, b: B, c: C, callback: Nodeback<L, R>) => void, a: A, b: B, c: C): Future<L, R>
+  export function encaseN3<L, R, A, B, C>(fn: (a: A, b: B, c: C, callback: Nodeback<L, R>) => void, a: A, b: B): (c: C) => Future<L, R>
+  export function encaseN3<L, R, A, B, C>(fn: (a: A, b: B, c: C, callback: Nodeback<L, R>) => void, a: A): AwaitingTwo<B, C, Future<L, R>>
+  export function encaseN3<L, R, A, B, C>(fn: (a: A, b: B, c: C, callback: Nodeback<L, R>) => void): AwaitingThree<A, B, C, Future<L, R>>
+
+  /** Encase the given Promise-returning function such that it returns a Future of its resolution value. See https://github.com/fluture-js/Fluture#encasep */
+  export function encaseP<L, R, A>(fn: (a: A) => Promise<R>, a: A): Future<L, R>
+  export function encaseP<L, R, A>(fn: (a: A) => Promise<R>): (a: A) => Future<L, R>
+
+  /** Encase the given Promise-returning function such that it returns a Future of its resolution value. See https://github.com/fluture-js/Fluture#encasep */
+  export function encaseP2<L, R, A, B>(fn: (a: A, b: B) => Promise<R>, a: A, b: B): Future<L, R>
+  export function encaseP2<L, R, A, B>(fn: (a: A, b: B) => Promise<R>, a: A): (b: B) => Future<L, R>
+  export function encaseP2<L, R, A, B>(fn: (a: A, b: B) => Promise<R>): AwaitingTwo<A, B, Future<L, R>>
+
+  /** Encase the given Promise-returning function such that it returns a Future of its resolution value. See https://github.com/fluture-js/Fluture#encasep */
+  export function encaseP3<L, R, A, B, C>(fn: (a: A, b: B, c: C) => Promise<R>, a: A, b: B, c: C): Future<L, R>
+  export function encaseP3<L, R, A, B, C>(fn: (a: A, b: B, c: C) => Promise<R>, a: A, b: B): (c: C) => Future<L, R>
+  export function encaseP3<L, R, A, B, C>(fn: (a: A, b: B, c: C) => Promise<R>, a: A): AwaitingTwo<B, C, Future<L, R>>
+  export function encaseP3<L, R, A, B, C>(fn: (a: A, b: B, c: C) => Promise<R>): AwaitingThree<A, B, C, Future<L, R>>
+
+  /** Extract the rejection reason into an Array if it's present. See https://github.com/fluture-js/Fluture#extractleft */
+  export function extractLeft<L, R>(source: Future<L, R>): Array<L>
+
+  /** Extract the resolution value into an Array if it's present. See https://github.com/fluture-js/Fluture#extractright */
+  export function extractRight<L, R>(source: Future<L, R>): Array<R>
+
+  /** Fold both branches into the resolution branch. See https://github.com/fluture-js/Fluture#fold */
+  export function fold<LA, RA, RB>(lmapper: (left: LA) => RA, rmapper: (right: RA) => RB, source: Future<LA, RA>): Future<never, RB>
+  export function fold<LA, RA, RB>(lmapper: (left: LA) => RA, rmapper: (right: RA) => RB): (source: Future<LA, RA>) => Future<never, RB>
+  export function fold<LA, RA, RB>(lmapper: (left: LA) => RA): AwaitingTwo<(right: RA) => RB, Future<LA, RA>, Future<never, RB>>
+
+  /** Fork the given Future into the given continuations. See https://github.com/fluture-js/Fluture#fork */
+  export function fork<L, R>(reject: RejectFunction<L>, resolve: ResolveFunction<R>, source: Future<L, R>): Cancel
+  export function fork<L, R>(reject: RejectFunction<L>, resolve: ResolveFunction<R>): (source: Future<L, R>) => Cancel
+  export function fork<L, R>(reject: RejectFunction<L>): AwaitingTwo<ResolveFunction<R>, Future<L, R>, Cancel>
+
+  /** Build a coroutine using Futures. See https://github.com/fluture-js/Fluture#go */
+  export function go<L, R>(generator: Generator<Future<L, any>, R>): Future<L, R>
+
+  /** Manage resources before and after the computation that needs them. See https://github.com/fluture-js/Fluture#hook */
+  export function hook<L, H, R>(acquire: Future<L, H>, dispose: (handle: H) => Future<L, any>, consume: (handle: H) => Future<L, R>): Future<L, R>
+  export function hook<L, H, R>(acquire: Future<L, H>, dispose: (handle: H) => Future<L, any>): (consume: (handle: H) => Future<L, R>) => Future<L, R>
+  export function hook<L, H, R>(acquire: Future<L, H>): AwaitingTwo<(handle: H) => Future<L, any>, (handle: H) => Future<L, R>, Future<L, R>>
+
+  /** Returns true for Futures. See https://github.com/fluture-js/Fluture#isfuture */
+  export function isFuture(value: any): boolean
+
+  /** Returns true for Futures that will certainly never settle. See https://github.com/fluture-js/Fluture#isnever */
+  export function isNever(value: any): boolean
+
+  /** Set up a cleanup Future to run after the given action has settled. See https://github.com/fluture-js/Fluture#lastly */
+  export function lastly<L, R>(cleanup: Future<L, any>, action: Future<L, R>): Future<L, R>
+  export function lastly<L, R>(cleanup: Future<L, any>): (action: Future<L, R>) => Future<L, R>
+
+  /** Map over the resolution value of the given Future. See https://github.com/fluture-js/Fluture#map */
+  export function map<L, RA, RB>(mapper: (value: RA) => RB, source: Future<L, RA>): Future<L, RB>
+  export function map<L, RA, RB>(mapper: (value: RA) => RB): (source: Future<L, RA>) => Future<L, RB>
+
+  /** Map over the rejection reason of the given Future. See https://github.com/fluture-js/Fluture#maprej */
+  export function mapRej<LA, LB, R>(mapper: (reason: LA) => LB, source: Future<LA, R>): Future<LB, R>
+  export function mapRej<LA, LB, R>(mapper: (reason: LA) => LB): (source: Future<LA, R>) => Future<LB, R>
+
+  /** A Future that never settles. See https://github.com/fluture-js/Fluture#never */
+  export const never: Future<never, never>
+
+  /** Create a Future using a provided Node-style callback. See https://github.com/fluture-js/Fluture#node */
+  export function node<L, R>(fn: (done: Nodeback<L, R>) => void): Future<L, R>
+
+  /** Create a Future with the given resolution value. See https://github.com/fluture-js/Fluture#of */
+  export function of<R>(value: R): Future<never, R>
+
+  /** Logical or for Futures. See https://github.com/fluture-js/Fluture#or */
+  export function or<L, R>(left: Future<L, R>, right: Future<L, R>): Future<L, R>
+  export function or<L, R>(left: Future<L, R>): (right: Future<L, R>) => Future<L, R>
+
+  /** Run an Array of Futures in parallel, under the given concurrency limit. See https://github.com/fluture-js/Fluture#parallel */
+  export function parallel<L, R>(concurrency: number, futures: Array<Future<L, R>>): Future<L, Array<R>>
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -3,6 +3,7 @@
   "version": "6.2.8",
   "description": "FantasyLand compliant (monadic) alternative to Promises",
   "main": "index.js",
+  "typings": "index.d.ts",
```
davidchambers: It seems that `"typings"` is an alias for `"types"`. Let's use `"types"` instead.

tusharmath: @davidchambers Done.

index.d.ts
```diff
@@ -0,0 +1,130 @@
...
+    (value: T): void
+  }
+  interface RejectFunction {
+    (error: Error): void
+  }
+
+  interface FunctionReturning<T extends Function> {
+    (...t: any[]): T
+  }
+  abstract class Future<T> {
```
Avaq: Can we define `Future<L, R>` instead, in order to define restrictions for the "error" branch?

index.d.ts
```diff
@@ -0,0 +1,130 @@
...
+    done<T> (fn: (err: Error, value: T) => void): Future<T>
```
Avaq: 1. We better specify `fork` over `done`: `done` is really just an alias.
1. I'm not sure, but this seems to state that `done` returns a Future, instead it returns `Cancel`. Maybe we can define a type for `Cancel`, as is [done in the README](https://github.com/fluture-js/Fluture#types).

index.d.ts
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  function after<T> (duration: number, value: string): Future<T>
```
Avaq: Would it be possible to define a `PositiveInteger` and use that to restrict the input for `duration`?

index.d.ts
```diff
@@ -0,0 +1,130 @@
...
+
+  function and<A, B> (a: Future<A>, b: Future<B>): Future<A | B>
```
Avaq: The return type can be specialized to be `Future<B>` (the resolution value of `a` is always ignored), when we introduce the error branch I think we can further improve the signature to:

```ts
function and<L, R> (a: Future<L, any>, b: Future<L, R>): Future<L, R>
```

index.d.ts
```diff
@@ -0,0 +1,130 @@
...
+
+  // TODO: Need Help
+  function never<T> (): Future<T>
+
+  // TODO: Need Help
+  function node<T> (): Future<T>
+
+  function of<T> (value: T): Future<T>
+
+  function or<A, B> (a: Future<A>, b: Future<B>): Future<A | B>
```
Avaq: Although it's *technically* possible to provide two futures of different types to this function, I consider this a flaw in JavaScript due to the lack of its type system. Now that we *have* a type system, I think we should guide users to use consistent types. Therefore, I would like to see this signature change to something like:

```ts
function or<L, R> (a: Future<L, R>, b: Future<L, R>): Future<L, R>
```

index.d.ts
```diff
@@ -0,0 +1,130 @@
...
+
+  // TODO: Need Help
+  function par<T> (): Future<T>
+
+  function parallel<T> (concurrency: number, futures: Array<Future<any>>): Future<Array<T>>
```
Avaq: This `any` there could be `T`, right?

Avaq:
Thank you for the effort thus far @tusharmath! I'm excited about providing this feature to users. Many of my colleagues rely on TypeScript for autocompletions and other goodies, and I would be delighted if these are available for Fluture as well!

index.d.ts
```diff
@@ -0,0 +1,128 @@
...
+    (error: S): void
+  }
+
+  interface FunctionReturning<T> {
+    (...t: any[]): T
+  }
+  abstract class Future<T, S> {
+    done<T, S> (fn: (err: S, value: T) => void): Future<T, S>
+
+    fork (reject: RejectFunction<S>, resolve: ResolveFunction<T>): void
```
Avaq: [`fork`](https://github.com/fluture-js/Fluture#fork) returns `Cancel`

index.d.ts
```diff
@@ -0,0 +1,128 @@
...
+  }
+  interface RejectFunction<S> {
+    (error: S): void
+  }
+
+  interface FunctionReturning<T> {
+    (...t: any[]): T
+  }
+  abstract class Future<T, S> {
+    done<T, S> (fn: (err: S, value: T) => void): Future<T, S>
```
Avaq: [`done`](https://github.com/fluture-js/Fluture#done) returns `cancel`

index.d.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    fork (reject: RejectFunction<S>, resolve: ResolveFunction<T>): void
+  }
```
Avaq: Fluture has many more methods than just `fork` and `done`. Almost every function in Fluture is also available as a method. We should provide typings for these as well.

index.d.ts
```diff
@@ -0,0 +1,128 @@
+declare module 'fluture' {
+  interface ResolveFunction<T> {
+    (value: T): void
+  }
+  interface RejectFunction<S> {
+    (error: S): void
+  }
+
+  interface FunctionReturning<T> {
+    (...t: any[]): T
```
Avaq: Fluture has no variadic functions in its code-base. I think we should be able to avoid using this type.

index.d.ts
```diff
@@ -0,0 +1,128 @@
...
+  }
+  abstract class Future<T, S> {
+    done<T, S> (fn: (err: S, value: T) => void): Future<T, S>
+
+    fork (reject: RejectFunction<S>, resolve: ResolveFunction<T>): void
+  }
+
+  function after<T, S> (duration: number, value: string): Future<T, S>
```
Avaq: Can we define a `PositiveInteger` type?

index.d.ts
```diff
@@ -0,0 +1,128 @@
...
+    (value: T): void
+  }
+  interface RejectFunction<S> {
+    (error: S): void
+  }
+
+  interface FunctionReturning<T> {
+    (...t: any[]): T
+  }
+  abstract class Future<T, S> {
```
Avaq: I would prefer that the positions of the generic types correspond to the positions of the functions in `fork`. In my example I named them `L` and `R` to mean "left" and "right" respectively. We can keep the names `S` and `T`, if that's a convention of sorts, but I would prefer the generic on the left to represent the type of the *rejection* reason, and the generic on the right to represent the *resolution* value.

index.d.ts
```diff
@@ -0,0 +1,128 @@
...
+    done<T, S> (fn: (err: S, value: T) => void): Future<T, S>
+
+    fork (reject: RejectFunction<S>, resolve: ResolveFunction<T>): void
+  }
+
+  function after<T, S> (duration: number, value: string): Future<T, S>
+
+  function and<A, B> (a: Future<A, any>, b: Future<B, any>): Future<A | B, any>
```
Avaq: I would like to avoid mixed types, and force users to use consistent types. We can avoid a mixed type in this function by making the input more strict:

```ts
function and<L, R> (a: Future<L, any>, b: Future<L, R>): Future<L, R>
```

index.d.ts
```diff
@@ -0,0 +1,128 @@
...
+
+  // TODO: Need Help
+  function ap<T, S> (): Future<T, S>
+
+  function attempt<T, S> (f: () => T): Future<T, S>
```
Avaq: Is it possible in TypeScript to capture the type of the values that may be thrown from a function, similarly to how Java has `void fn() throws TypeA, TypeB {...}`?

index.d.ts
```diff
@@ -0,0 +1,128 @@
...
+
+  // TODO: Need Help
+  function never<T, S> (): Future<T, S>
+
+  // TODO: Need Help
+  function node<T, S> (): Future<T, S>
+
+  function of<T, S> (value: T): Future<T, S>
+
+  function or<A, B> (a: Future<A, any>, b: Future<B, any>): Future<A | B, any>
```
Avaq: See my comment under `and`

index.d.ts
tusharmath: I don't think thats possible using just typescript.

index.d.ts
tusharmath: Do you think it would make more sense to have something like —
```ts
or<L1, R1, L2, R2> (a: Future<L1, R1>, b: Future<L2, R2>): Future<L1|L2, R1|R2>
```

index.d.ts
tusharmath: Well this is an internal type I can make it private to this module.

index.d.ts
tusharmath: I thought we are only going to add typings for `fork` ? 
https://github.com/fluture-js/Fluture/pull/129#discussion_r127650946

index.d.ts
tusharmath: May be once we are done with the static functions we could update the methods also?

index.d.ts
tusharmath: I don't think thats possible in Typescript as of today.

index.d.ts
```diff
@@ -0,0 +1,133 @@
+declare module 'fluture' {
+  interface ResolveFunction<T> {
```
Avaq: Shall we rename the generic to `R` for consistency?

index.d.ts
```diff
@@ -0,0 +1,133 @@
...
+    (value: T): void
+  }
+  interface RejectFunction<S> {
```
Avaq: Shall we rename the generic to `L` for consistency?

index.d.ts
```diff
@@ -0,0 +1,133 @@
...
+
+  interface Cancel {
+    (): void
+  }
+
+  export abstract class IFuture<L, R> {
+    fork (reject: RejectFunction<L>, resolve: ResolveFunction<R>): Cancel
+  }
+
+  export function after<L, R> (duration: number, value: string): IFuture<L, R>
```
Avaq: The `value` should be of type `R`.

index.d.ts
```diff
@@ -0,0 +1,133 @@
...
+
+  export function and<L, R> (a: IFuture<L, R>, b: IFuture<L, R>): IFuture<L, R>
```
Avaq: We can relax this signature a little bit by using allowing `a` to be a `IFuture<L, any>`, since the `R` of `a` is always ignored. Let's also rename `a` to `left` and `b` to `right`.

index.d.ts
```diff
@@ -0,0 +1,133 @@
...
+
+  // TODO: Need Help
+  export function ap<L, R> (): IFuture<L, R>
+
+  export function attempt<L, R> (f: () => L): IFuture<L, R>
+
+  export function bimap<A, B, AA, BB> (left: (a: A) => AA,
+                                       right: (a: B) => BB): IFuture<AA, any>
+
+  export function both<A, B> (a: IFuture<A, any>, b: IFuture<B, any>): IFuture<[A, B], any>
```
Avaq: I think this one needs to be updated to use the swapped positions of the generics:

```ts
export function both<L, A, B> (left: IFuture<L, A>, right: IFuture<L, B>): IFuture<L, [A, B]>
```

index.d.ts
```diff
@@ -0,0 +1,133 @@
...
+
+  export function cache<L, R> (f: IFuture<L, any>): IFuture<L, R>
```
Avaq: Since the resolution value of a cached Future is the same as the resolution value of the input Future, we can replace this `any` by `R`. We should also rename `f` to `source`.

index.d.ts
```diff
@@ -0,0 +1,133 @@
...
+
+  export function chain<L, R> (fn: (s: R) => IFuture<IFuture<L, any>, any>, src: IFuture<R, any>): IFuture<L, R>
```
Avaq: This needs several improvements:

* The return value from `fn` doesn't need to be a higher order Future.
* I think you missed some swaps of generics.
* We need to introduce a third generic to capture the transformation.
* Let's use valid English words as argument identifiers where we can.

```ts
export function chain<L, A, B> (mapper: (value: A) => IFuture<L, B>, source: IFuture<L, A>): IFuture<L, B>
```

index.d.ts
```diff
@@ -0,0 +1,133 @@
...
+
+  export function isFuture<L, R> (fut: any): boolean
+
+  // TODO: Need Help
+  export function isNever<L, R> (): boolean
+
+  // TODO: Need Help
+  export function lastly<L, R> (): IFuture<L, R>
+
+  export function map<L, R, M> (fn: (value: L) => M, future: IFuture<L, R>): IFuture<M, R>
```
Avaq: I think you missed a swap of generics here:

```ts
export function map<L, A, B> (mapper: (value: A) => B, source: IFuture<L, A>): IFuture<L, B>
```

index.d.ts
```diff
@@ -0,0 +1,133 @@
...
+  export function node<L, R> (): IFuture<L, R>
+
+  export function of<L, R> (value: L): IFuture<L, R>
+
+  export function or<L, R> (a: IFuture<L, R>, b: IFuture<L, R>): IFuture<L, R>
+
+  // TODO: Need Help
+  export function par<L, R> (): IFuture<L, R>
+
+  export function parallel<L, R> (concurrency: number, futures: Array<IFuture<any, any>>): IFuture<Array<L>, any>
```
Avaq: Is it not possible to force all Futures to be of the same type?

```ts
export function parallel<L, R> (concurrency: Integer, futures: Array<IFuture<L, R>>): IFuture<L, Array<R>>
```

index.d.ts
```diff
@@ -0,0 +1,133 @@
...
+
+  export function promise<L> (f: IFuture<L, any>): Promise<L>
```
Avaq: I'm not sure where the Promise type is coming from, or why it only has one generic, but let's swap our own generics:

```ts
export function promise<R> (f: IFuture<any, R>): Promise<R>
```

index.d.ts
```diff
@@ -0,0 +1,133 @@
...
+
+  export function race<A, B> (a: IFuture<A, any>, b: IFuture<B, any>): IFuture<A | B, any>
```
Avaq: Let's force users not to mix their types:

```ts
export function race<L, R> (a: IFuture<L, R>, b: IFuture<L, R>): IFuture<L, R>
```

index.d.ts
```diff
@@ -0,0 +1,133 @@
...
+  // TODO: Need Help
+  export function swap<L, R> (): IFuture<L, R>
+
+  // TODO: Need Help
+  export function tryP<L, R> (): IFuture<L, R>
+
+  // TODO: Need Help
+  export function value<L, R> (): IFuture<L, R>
+
+  export function Future<L, R> (fn: (reject: RejectFunction<L>, resolve: ResolveFunction<R>) => Cancel): IFuture<L, R>
```
Avaq: Let's rename `fn` to `computation`.

index.d.ts
```diff
@@ -0,0 +1,133 @@
...
+  export function cache<L, R> (f: IFuture<L, any>): IFuture<L, R>
+
+  export function chain<L, R> (fn: (s: R) => IFuture<IFuture<L, any>, any>, src: IFuture<R, any>): IFuture<L, R>
+
+  export function encase2<L, R, A, B> (fn: (a: A, b: B) => L, a: A, b: B): IFuture<L, R>
+
+  export function encase3<L, R, A, B, C> (fn: (a: A, b: B, c: C) => L, a: A, b: B, c: C): IFuture<L, R>
+
+  export function encase<L, R> (fn: FunctionReturning<L>, ...t: any[]): IFuture<L, R>
+  export function encase<L, R> (fn: FunctionReturning<L>): {(...t: any[]): IFuture<L, R>}
```
Avaq: Why do you use `FunctionReturning<T>` here, and `(...) => T` in the other arities of `encase`? I would prefer the latter for all arities, and removing the `FunctionReturning` type.

index.d.ts
```diff
@@ -0,0 +1,133 @@
...
+  // TODO: Need Help
+  export function fork<L, R> (): IFuture<L, R>
+
+  // TODO: Need Help
+  export function go<L, R> (): IFuture<L, R>
+
+  // TODO: Need Help
+  export function hook<L, R> (): IFuture<L, R>
+
+  export function isFuture<L, R> (fut: any): boolean
```
Avaq: Let's rename `fut` to `value`.

index.d.ts
tusharmath: @Avaq What would happen if the mapper rejects with a value which isn't of type `L`?

index.d.ts
tusharmath: Again I realized this earlier today :) 
does the following make sense —
```ts
export function encase<L, R> (fn: () => R): {(): IFuture<L, R>}
export function encase<L, R, V> (fn: (v: V) => R, t: V): IFuture<L, R>
export function encase<L, R, V> (fn: (v: V) => R): {(t: V): IFuture<L, R>}
```

Apparently `encase` is polymorphic so we should support all the three cases.

index.d.ts
tusharmath: I would personally prefer keeping the same semantics and assume that the source always  is represented as — `IFuture<L, R>`

```
export function map<L, R, M> (fn: (value: L) => M, source: IFuture<L, R>): IFuture<M, R>
export function map<L, R, M> (fn: (value: L) => M): ({(source: IFuture<L, R>): IFuture<M, R>})
```

index.d.ts
tusharmath: I think its better to keep it this way because different types can exist. Like making multiple HTTP requests all of them returning different json responses. Developers this way can specify the types manually for each future and the combinator functions.

index.d.ts
tusharmath: Promises are available globally out of the box if you are using typescript in your project.

index.d.ts
```diff
@@ -0,0 +1,170 @@
...
+    apply: IFuture<L, (value: VR) => R>,
+    value: IFuture<L, VR>
+  ): IFuture<L, R>
+
+  export function attempt<L, R>(f: () => L): IFuture<L, R>
+
+  export function bimap<A, B, AA, BB>(
+    left: (a: A) => AA,
+    right: (a: B) => BB
+  ): IFuture<AA, any>
```
Avaq: Let's use `LA`, `LB`, `RA`, and `RB`. I think it's also nice to use names that correspond to the variable names used in Fluture's source-code. Furthermore, this function is missing its third parameter.

```ts
bimap<LA, LB, RA, RB>(
  lmapper: (reason: LA) => LB),
  rmapper: (value: RA) => RB),
  source: IFuture<LA, RA>
): IFuture<LB, RB>
```

index.d.ts
```diff
@@ -0,0 +1,170 @@
...
+
+  export function both<L, A, B>(
+    left: IFuture<L, A>,
+    right: IFuture<L, B>
+  ): IFuture<L, [A, B]>
+
+  export function cache<L, R>(source: IFuture<L, R>): IFuture<L, R>
+
+  export function chain<L, R, RR>(
```
Avaq: Shall we use `RA` and `RB`?

index.d.ts
```diff
@@ -0,0 +1,170 @@
...
+    mapper: (value: R) => IFuture<L, RR>,
+    source: IFuture<L, R>
+  ): IFuture<L, RR>
+
+  export function done<L, R>(
+    fn: (cb: (err: L | null, value: R) => void) => void
+  ): IFuture<L, R>
```
Avaq: It seems you confused this function for `Future.node`. Instead it be somewhat like this:

```ts
done<L, R>(
  callback: (err: L | null, value: R?) => void,
  source: IFuture<L, R>
): Cancel
```

index.d.ts
```diff
@@ -0,0 +1,170 @@
...
+
+  export function encase2<L, R, A, B>(
+    fn: (a: A, b: B) => L,
```
Avaq: The return value of the function should be `R`, not `L`.

index.d.ts
```diff
@@ -0,0 +1,170 @@
...
+    left: IFuture<L, any>,
+    right: IFuture<L, R>
+  ): IFuture<L, R>
+
+  export function ap<L, R, VR>(
+    apply: IFuture<L, (value: VR) => R>,
+    value: IFuture<L, VR>
+  ): IFuture<L, R>
+
+  export function attempt<L, R>(f: () => L): IFuture<L, R>
```
Avaq: The return value of the function should be `R`, not `L`.

index.d.ts
```diff
@@ -0,0 +1,170 @@
...
+  ): IFuture<L, R>
+
+  export function encase2<L, R, A, B>(
+    fn: (a: A, b: B) => L,
+    a: A,
+    b: B
+  ): IFuture<L, R>
+
+  export function encase3<L, R, A, B, C>(
+    fn: (a: A, b: B, c: C) => L,
```
Avaq: The return value of the function should be `R`, not `L`.

index.d.ts
```diff
@@ -0,0 +1,170 @@
...
+    a: A,
+    b: B,
+    c: C
+  ): IFuture<L, R>
+
+  export function encase<L, R>(fn: () => R): {(): IFuture<L, R>}
+  export function encase<L, R, V>(fn: (v: V) => R, t: V): IFuture<L, R>
+  export function encase<L, R, V>(fn: (v: V) => R): {(t: V): IFuture<L, R>}
```
Avaq: Why is `encase` overloaded? Is this the way to express currying?

index.d.ts
```diff
@@ -0,0 +1,170 @@
...
+  // TODO: Need Help
+  export function fork<L, R>(): IFuture<L, R>
+
+  // TODO: Need Help
+  export function go<L, R>(): IFuture<L, R>
+
+  // TODO: Need Help
+  export function hook<L, R>(): IFuture<L, R>
+
+  export function isFuture<L, R>(value: any): boolean
```
Avaq: No need for the reservation of the generics `L` and `R` here.

index.d.ts
```diff
@@ -0,0 +1,170 @@
...
+
+  // TODO: Need Help
+  export function isNever<L, R>(): boolean
```
Avaq: This has the same signature as `isFuture`.

index.d.ts
```diff
@@ -0,0 +1,170 @@
...
+
+  // TODO: Need Help
+  export function lastly<L, R>(): IFuture<L, R>
+
+  export function map<L, R, M>(
```
Avaq: Let's use `RA` and `RB`.

index.d.ts
```diff
@@ -0,0 +1,170 @@
...
+
+  // TODO: Need Help
+  export function never<L, R>(): IFuture<L, R>
+
+  // TODO: Need Help
+  export function node<L, R>(): IFuture<L, R>
+
+  export function of<L, R>(value: L): IFuture<L, R>
+
+  export function or<L, R>(a: IFuture<L, R>, b: IFuture<L, R>): IFuture<L, R>
```
Avaq: Let's use `left` and `right` as parameter names.

index.d.ts
```diff
@@ -0,0 +1,170 @@
...
+
+  // TODO: Need Help
+  export function par<L, R>(): IFuture<L, R>
+
+  export function parallel<L, R>(
+    concurrency: number,
+    futures: Array<IFuture<any, any>>
+  ): IFuture<L, Array<any>>
```
Avaq: I *really* prefer to force developers to use consistent types. Again, the fact that it's *possible* in Fluture is simply because it has a huge performance impact to impose this restriction at run-time.

> Like making multiple HTTP requests all of them returning different json responses

I would like for Fluture to move developers to normalize the different responses before passing the futures into `parallel`. Futures that resolve to different types of values should have never made it into the same list.

The same goes for the rejection branch. Once you end up with a Future of a mixed type in your rejection branch, you're going to have a bad time handling that in `fork`. My preference is still for changing this signature to:

```ts
parallel<L, R>(
  concurrency: number,
  futures: Array<IFuture<L, R>>
): IFuture<L, Array<R>>
```

index.d.ts
```diff
@@ -0,0 +1,170 @@
...
+
+  export function promise<R>(f: IFuture<any, R>): Promise<R>
```
Avaq: Let's rename `f` to `source`.

index.d.ts
```diff
@@ -0,0 +1,170 @@
+declare module 'fluture' {
+  interface ResolveFunction<R> {
+    (value: R): void
+  }
+  interface RejectFunction<L> {
+    (error: L): void
```
Avaq: Let's rename `error` to `reason`.

index.d.ts
```diff
@@ -0,0 +1,170 @@
...
+  ): IFuture<L, Array<any>>
+
+  export function promise<R>(f: IFuture<any, R>): Promise<R>
+
+  export function race<L, R>(
+    left: IFuture<L, R>,
+    right: IFuture<L, R>
+  ): IFuture<L, R>
+
+  export function reject<L, R>(value: L): IFuture<L, R>
```
Avaq: Let's rename `value` to `reason`.

tusharmath:
@Avaq I guess because of different time zones we the to and fro take long. To save time I have added you as a collaborator to my fork.

https://github.com/tusharmath/Fluture/invitations

index.d.ts
```diff
@@ -0,0 +1,170 @@
...
+
+  export function after<L, R>(duration: number, value: R): IFuture<L, R>
+
+  export function and<L, R>(
+    left: IFuture<L, any>,
+    right: IFuture<L, R>
+  ): IFuture<L, R>
+
+  export function ap<L, R, VR>(
+    apply: IFuture<L, (value: VR) => R>,
```
gcanti: > the ap function (as opposed to the method) actually works on any Apply

Now I'm curious, what do you mean?

Avaq: Let's have a look at the [documentation for `ap`](https://github.com/fluture-js/Fluture#ap). When you expand the function signature, we see this:

```hs
ap                  :: Apply m => m        (a -> b) -> m        a -> m        b
Future.ap           :: Apply m => m        (a -> b) -> m        a -> m        b
Future.prototype.ap ::            Future e (a -> b) ~> Future e a -> Future e b
```

This shows us that there are *three* distinct versions of the `ap` function. The first and the second are the same, so we can ignore that. It just serves to show that the function lives on the `Future` class *as well as* as a named export (for Static Land compliance).

Let's compare the first and the third.

**The first** is the named export. It's a function with signature:

```hs
ap :: m (a -> b) -> m a -> m b
```

Or in TypeScript:

```ts
function ap<T, A, B>((A) => B, T<A>): T<B>
```

Except that the Hindley-Milner signature has one more part, the leading `Apply m => `. This indicates that, though `m` is a generic type, it must conform to the [`Apply` interface (as defined by Fantasy Land)](https://github.com/fantasyland/fantasy-land#apply).

**The third** is the `ap` *method*. It lives on the prototype of the Future class and is essentially `ap`, but specialized for the Future type:

```hs
ap :: Future e (a -> b) ~> Future e a -> Future e b
```

The `m` has been replaced with `Future e`, and the `Apply m =>` restriction can therefore be omitted. I find it difficult to translate this to TypeScript, let me illustrate:

```ts
class Future<L, R> {
  ap<A, B>(value: Future<L, A>): Future<L, B>
}
```

This TypeScript definition is pretty crummy, as it fails to show where `A` goes, and where `B` comes from. The fact being that `R` in this case must be `(A) => B`. If we would want to improve it, we would have to make use of interfaces such as is done [here](https://github.com/gcanti/fp-ts/blob/master/src/Apply.ts#L17-L19), I think.

gcanti: Ah got it, thanks. So `Future.ap` is not specialised for the Future type (and it leverages `Z.ap` I guess)

gcanti: My suggestion would be to specialise `Future.ap` too, since using a particular encoding for HKTs would couple this library with that implementation. 

Writing the bindings of `Fluture` for `fp-ts` is actually pretty easy and could be done in userland

```ts
//
// fp-ts bindings for Fluture
//

import { ap, of, map, chain, IFuture } from 'fluture'
import { Monad } from 'fp-ts/lib/Monad'

declare module 'fp-ts/lib/HKT' {
  interface URI2HKT2<L, A> {
    Fluture: IFuture<L, A>
  }
}

declare module 'fluture' {
  interface IFuture<L, R> {
    _A: R
    _L: L
    _URI: URI
  }
}

export const URI = 'Fluture'

export type URI = typeof URI

export const monad: Monad<URI> = {
  URI,
  map,
  of,
  ap,
  chain
}

//
// usage
//

import { lift } from 'fp-ts/lib/Functor'

const double = (n: number) => n * 2

// liftedDouble: <L>(fa: IFuture<L, number>) => IFuture<L, number>
export const liftedDouble = lift(monad, double)
```

index.d.ts
```diff
@@ -1,170 +1,167 @@
...
+
+  }
+
+  interface Future<L, R> {
     fork(reject: RejectFunction<L>, resolve: ResolveFunction<R>): Cancel
   }
 
-  export function after<L, R>(duration: number, value: R): IFuture<L, R>
+  export function after<L, R>(duration: number, value: R): Future<L, R>
+  export function after<L, R>(duration: number): ({(value: R): Future<L, R>})
```
gcanti: nitpick: the form

```ts
export function after<L, R>(duration: number): (value: R) => Future<L, R>
```

is more idiomatic

Avaq: Please nitpick :)
That's a great improvement!

index.d.ts
```diff
@@ -31,75 +31,75 @@ declare module 'fluture' {
   }
 
   export function after<L, R>(duration: number, value: R): Future<L, R>
```
gcanti: You can add some documentation here

```ts
  /** Creates a Future which resolves with the given value after n milliseconds. */
  export function after<L, R>(duration: number, value: R): Future<L, R>
  export function after<L, R>(duration: number): (value: R) => Future<L, R>
```

will show up in the IDE (for example vscode)

Avaq: Thanks for the tip. Will do! :)

Avaq:
Right. Onward to the methods:

```text
[fluture]$ cat README.md | grep -E 'prototype.*::' | sed -r 's/ +/ /g'
Future.prototype.map :: Future a b ~> (b -> c) -> Future a c
Future.prototype.map :: Future e a ~> (a -> b) -> Future e b
Future.prototype.bimap :: Future a b ~> (a -> c) -> (b -> d) -> Future c d
Future.prototype.chain :: Future e a ~> (a -> Future e b) -> Future e b
Future.prototype.swap :: Future a b ~> Future b a
Future.prototype.mapRej :: Future a b ~> (a -> c) -> Future c b
Future.prototype.chainRej :: Future a b ~> (a -> Future a c) -> Future a c
Future.prototype.fold :: Future a b ~> (a -> c, b -> c) -> Future d c
Future.prototype.ap :: Future e (a -> b) ~> Future e a -> Future e b
Future.prototype.and :: Future a b ~> Future a c -> Future a c
Future.prototype.or :: Future a b ~> Future a b -> Future a b
Future.prototype.fork :: Future a b ~> (a -> x, b -> x) -> Cancel
Future.prototype.value :: Future a b ~> (b -> x) -> Cancel
Future.prototype.done :: Future a b ~> Nodeback a b -> Cancel
Future.prototype.promise :: Future a b ~> Promise b a
Future.prototype.race :: Future a b ~> Future a b -> Future a b
Future.prototype.both :: Future a b ~> Future a c -> Future a (Pair b c)
Future.prototype.finally :: Future a b ~> Future a c -> Future a b
Future.prototype.lastly :: Future a b ~> Future a c -> Future a b
```

Avaq:
@tusharmath I added all methods, all remaining functions, currying overloadings to all functions and documentation to all methods and functions. Let's wait for a review of @gcanti and then I'll ask you to rebase on `fluture-js:master` and squash all commits.

index.d.ts
```diff
@@ -0,0 +1,299 @@
...
+  export function encase2<L, R, A, B>(fn: (a: A, b: B) => R): (a: A) => (b: B) => Future<L, R>
+
+  /* Encase the given function such that it returns a Future of its return value. See https://github.com/fluture-js/Fluture#encase */
+  export function encase3<L, R, A, B, C>(fn: (a: A, b: B, c: C) => R, a: A, b: B, c: C): Future<L, R>
+  export function encase3<L, R, A, B, C>(fn: (a: A, b: B, c: C) => R, a: A, b: B): (c: C) => Future<L, R>
+  export function encase3<L, R, A, B, C>(fn: (a: A, b: B, c: C) => R, a: A): (b: B, c: C) => Future<L, R>
+  export function encase3<L, R, A, B, C>(fn: (a: A, b: B, c: C) => R, a: A): (b: B) => (c: C) => Future<L, R>
+  export function encase3<L, R, A, B, C>(fn: (a: A, b: B, c: C) => R): (a: A, b: B, c: C) => Future<L, R>
+  export function encase3<L, R, A, B, C>(fn: (a: A, b: B, c: C) => R): (a: A) => (b: B, c: C) => Future<L, R>
+  export function encase3<L, R, A, B, C>(fn: (a: A, b: B, c: C) => R): (a: A) => (b: B) => (c: C) => Future<L, R>
```
Avaq: This approach to currying doesn't actually seem to work:

```ts
import {encase3} from 'fluture'

encase3((a, b, c) => a + b + c) /*curry*/ ('a', 'b', 'c')
```
```text
                     ^^^^^
        Cannot use operator '+' on type '{}'
```

Or, in other words: TypeScript failed to infer that we are passing a String there, and `+` should be fine.

Avaq: Actually, when annotating `a:string, b:string, c:string`, TypeScript will let it pass.

Unfortunately there's another, bigger, problem. The return value of these functions is not overloaded. TypeScript considers the following two definitions a "conflict" because they both have the exact same types of arguments, but "differing" return values:

```ts
export function encase3<L, R, A, B, C>(fn: (a: A, b: B, c: C) => R): (a: A, b: B, c: C) => Future<L, R>
export function encase3<L, R, A, B, C>(fn: (a: A, b: B, c: C) => R): (a: A) => (b: B, c: C) => Future<L, R>
```

And it deals with it by simply ignoring the second.

Avaq:
I'm going on a two weeks vacation, starting in an hour. I'm fairly sure I won't be working on anything during this period, so unfortunately I will have to leave it at this.

If any of you may feel like continuing this work, @tusharmath could add you as a contributor to his fork. :)

I'll pick things back up once I have returned.

tusharmath:
I have been on a vacation my self :)
Will take a look at the PR in the evening today (IST).

Avaq:
I believe that all that's left to do is to rewrite all curried functions using the Partial interfaces, as shown in my last commit. :)

Avaq:
I completed the work, I rebased on `avaq:master`, and I squashed all commits made on this branch. I'm going to test this branch tomorrow in a TS environment to verify if things work fine. I believe these typings will only work for the ES6 *module* version of Fluture, correct? We might consider adding a note about this to the README.

Avaq:
I believe these typings will only work for the ES6 *module* version of Fluture, correct? We might consider adding a note about this to the README if so.

Furthermore, @gcanti told me it would be easy for users to include overloadings for polymorphic functions such as `map`, which are now typed as if they are specialized in `Future`. Perhaps a guiding text on how to accomplish this could also be added to the README. Note that we should take currying into account if possible.

Finally, I'm wondering whether there is some kind of automated testing of the typings that we can do, perhaps just linting it for pitfalls. Anyone know of any such tools?

gcanti:
> Finally, I'm wondering whether there is some kind of automated testing of the typings that we can do, perhaps just linting it for pitfalls. Anyone know of any such tools?

I'm using typings-checker for that, you can see some examples in [typelevel-ts](https://github.com/gcanti/typelevel-ts/blob/92ab0b666d1be461e344988881f69ae08060aab3/package.json#L10) or [io-ts](https://github.com/gcanti/io-ts/blob/11ef27725622bb74c7eaefadfa166beb13393e79/typings-checker/index.ts)

gcanti:
FYI just tested this last version against fp-ts and seems to work pretty well

```ts
//
// fp-ts bindings for Fluture (likely will end up in a package like fp-ts-rxjs https://github.com/gcanti/fp-ts-rxjs)
//

export const URI = 'Fluture'

export type URI = typeof URI

declare module 'fp-ts/lib/HKT' {
  interface URI2HKT2<L, A> {
    Fluture: Future<L, A>
  }
}

declare module 'fluture' {
  interface Future<L, R> {
    _A: R
    _L: L
    _URI: URI
  }
}

//
// Usage
//

import { map, of, ap, chain, Future, reject } from 'fluture'
import { Monad } from 'fp-ts/lib/Monad'

export const fluture: Monad<URI> = {
  URI,
  map,
  of,
  ap,
  chain
}

import { sequence } from 'fp-ts/lib/Traversable'
import * as array from 'fp-ts/lib/Array'

sequence(fluture, array)([of(1), reject('ops')]).fork(() => console.error('error'), xs => console.log(xs)) // => "error"
sequence(fluture, array)([of(1), of(2)]).fork(() => console.error('error'), xs => console.log(xs)) // => [1, 2]
```

Avaq:
> I'm using typings-checker

Thanks for the reference! I also just found this https://github.com/sanctuary-js/sanctuary/issues/254#issuecomment-318912461

Avaq:
Darn. It looks like our way of handling curried functions was yet again problematic: https://github.com/sanctuary-js/sanctuary/pull/431#discussion_r130345580

Avaq:
Right. I'm ready to merge this PR. Can you given it a last look @tusharmath, @davidchambers and @gcanti?

README.md
```diff
@@ -58,7 +58,8 @@ getPackageName('package.json')
 
 ### In ES6 or newer environments
 
-The `package.json` sets a `module`-field for build-tools like [Rollup][].
+The `package.json` sets a `module`-field for build-tools like [Rollup][]. The
+module version also has Typescript definitions.
```
davidchambers: `s/Typescript/TypeScript/`

gcanti:
LGTM

index.d.ts
```diff
@@ -0,0 +1,320 @@
...
+    value: T
+  }
+
+  interface Iterator<N, D> {
+    next(value?: N): Next<N> | Done<D>
+  }
+
+  interface Generator<Y, R> {
+    (): Iterator<Y, R>
+  }
```
tusharmath: isn't `Iterator` and `Generator` already available inside `lib.d.es6.ts` ?

Avaq: I don't know. By the looks of that interface, `Next<A>` and `Done<B>` must be expressed as one `IteratorResult<A|B>`. I don't like that because it means we would have to relax the type signature for `chainRec`.

Avaq:
I've merged the PR. If you feel strongly about the Iterator interfaces, @tusharmath, feel free to open a separate issue or PR. :)

gcanti:
@Avaq great, let me know when you release this so I can work on the fp-ts bindings package

tusharmath:
@Avaq This is great!

metasansana:
I see there is a never type here. Typescript has a never type according to [here](https://basarat.gitbooks.io/typescript/docs/types/never.html)

Is this a conflict? `tsc` complains that this module does not export that type. 

What's the purpose of this fluture''s never type? I see it's used as the left type parameter on the return type of `of`.

Avaq:
As I understand, TypeScript allows *types* and *values* to share names, and the meaning of an identifier will change depending on context.

Fluture does not export a `never` *type*. It does export a *value* by the `never` identifier, defined [here](https://github.com/fluture-js/Fluture/blob/d4f493da639f70434bc569b1bcfa3b25e9b451e8/index.d.ts#L261).

The *type* of the `never` *value*, as well as the return types of `of`, `reject`, and `fold`, make use of the [built-in TypeScript `never` *type*](https://basarat.gitbooks.io/typescript/docs/types/never.html).

> `tsc` complains that this module does not export that type

You'll have to expand on that a little. When does it complain? If you're trying to use the `never` export from Fluture as a type, that might raise an error, because it is a value.

PR #131, opened 2017-07-18
Make the Parallel interpreter stack safe

### LINKED ISSUE #130: Max call stack exceeded on parallel(1, futures)
I expect `parallel` to be stack safe computation, but it's not. `futures` is about 9000 items length.

Is there way to calculate `futures` sequentially without using `chain` and `reduce` explicitly? What's the best practice in this case?

Avaq:
* Performance is unaffected.
* Added unit tests to combat the regression bugs I encountered while working on this.

Would you mind testing this branch against your code, @alexshuhin, to verify whether your issue has been resolved?

Fixes #130

### DIFF
--- a/test/5.parallel.test.js
+++ b/test/5.parallel.test.js
@@ -50,7 +50,7 @@ describe('Parallel', () => {
 
     it('parallelizes execution', function(){
       this.timeout(70);
-      const actual = parallel(2, [
+      const actual = parallel(5, [
         after(20, 'a'),
         after(20, 'b'),
         after(20, 'c'),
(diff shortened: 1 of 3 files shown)


test/5.parallel.test.js
```diff
@@ -50,7 +50,7 @@ describe('Parallel', () => {
 
     it('parallelizes execution', function(){
       this.timeout(70);
-      const actual = parallel(2, [
+      const actual = parallel(5, [
```
Avaq: This change is mostly unrelated. It corrects an earlier mistake which caused the tests to fail sometimes in Travis.

alexshuhin:
@Avaq WOW! it works perfectly! Amazing job :+1:, thx a lot

Performance is equals to my reduce/chain hack

Avaq:
Great! I will release it in `6.3.0` after #132 lands.

PR #132, opened 2017-07-18
Use duck-typing over other types of typing

Avaq:
This will allow Futures of incompatible versions, Futures spawned in other runtime-context (`vm.run*`), or even ["custom made" Futures](https://gitter.im/fluture-js/Fluture?at=596d0bd8f5b3458e30571d21) to be consumed by this function.

### DIFF
--- a/src/after.js
+++ b/src/after.js
@@ -8,7 +8,7 @@ function After$race(other){
        ? other
        : isNever(other)
        ? this
-       : (other instanceof After || other instanceof RejectAfter)
+       : typeof other._time === 'number'
        ? other._time < this._time ? other : this
        : Core._race.call(this, other);
 }
--- a/src/core.js
+++ b/src/core.js
@@ -329,7 +329,9 @@ Resolved.prototype.toString = function Resolved$toString(){
 
 export const of = x => new Resolved(x);
 
-function Never(){}
+function Never(){
+  this._isNever = true;
+}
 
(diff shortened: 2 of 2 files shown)


src/after.js
```diff
@@ -8,7 +8,7 @@ function After$race(other){
        ? other
        : isNever(other)
        ? this
-       : (other instanceof After || other instanceof RejectAfter)
+       : typeof other._time === 'number'
```
Avaq: Good old duck-typing.

src/core.js
```diff
@@ -329,7 +329,9 @@ Resolved.prototype.toString = function Resolved$toString(){
 
 export const of = x => new Resolved(x);
 
-function Never(){}
+function Never(){
+  this._isNever = true;
+}
```
Avaq: The reason I'm setting it in the constructor, rather than on the prototype, is that it will cause `inspect` to provide slightly more insightful input (`Future { _isNever: true }` vs `Future {}`).

src/after.js
davidchambers: Are these the only two occurrences of `instanceof` in the project?

Avaq: Technically, no. It's used in some places (like `isFuture`) to determine with high performance whether a value is what we want it to be, but these are the only occurrences left where there are no *or* conditions for the value to be identified as such.

PR #141, opened 2017-08-11
Add support for streams

tetsuo:
First of all, thank you guys! This really looks promising 👍 

@Avaq and I briefly talked about Fluture at WARP office this Wednesday. There was not enough time to discuss the details, but I couldn't help but immediately think of Streams! Searched the repository and couldn't find anything on it, so here we are :)

This looked like a really nice way to control streams as well, not sure how that would work tho. 
I have only implemented `tryS` yet, it doesn't consume anything, currently only concatenates the value, then resolves with it when stream ends or rejects if stream emits an error. What do you think? Any prior work you may recall?


tetsuo:
By the way I have only learned about Fluture, fantasyLand and such, less than 48 hours ago :) So, please consider this as a project to understand how this works :) 
I don't think this should be a core functionality at all, but I am also not quite sure about that :)

This can be incorporated as an external module as well, like "fluture-from-stream" or something.

Avaq:
Hello @tetsuo, thank you for your contribution! Nice to see that you are learning about our FP community!

> What do you think? Any prior work you may recall?
> This can be incorporated as an external module as well, like "fluture-from-stream" or something.

The functionality you've implemented is something I have considered adding to a separate module I would call `fluture-node`. The Fluture project currently has nothing built-in which would depend on a Node feature directly (the `encaseN` and `node` functions refer to the error-first *convention*, rather than a dependency on the environment). I would not want to bloat the code-base with functions certain to be useless in various environments.

I haven't gotten around to its creation, but the `fluture-node` module would contain helper functions for working with Stream, Buffer, EventEmitter, and possibly other Node-specific interfaces. For example:

```hs
awaitStream :: ReadableStream Buffer -> Future Error Buffer
once :: String -> EventEmitter -> Future Error Any
```

---

> please consider this as a project to understand how this works

How would you like to proceed? Would you prefer I gave a code review for you to learn more about Fluture?

tetsuo:
Hey @Avaq ! Yes, very interesting stuff, and this also appears to provide a nice and easy path to learn about the fundamentals of FP, thanks!

I agree, this functionality should be added as a module. A supplemental library like `fluture-node` sounds way better. Still, there is not much left until the new [Streams spec](https://streams.spec.whatwg.org/) lands in ;) Heard that it made to Chrome actually, but I have no idea yet.

No need for a code review. I'm just surfing around to see if I can replace a module where I've used  [multi-write-stream](https://github.com/mafintosh/multi-write-stream) with Fluture instead. I'll let you know about my findings soon!!

Closing this PR now, have a great weekend!

PR #144, opened 2017-08-18
Add RawGit distribution script

Avaq:
I tried to make a [Scrimba](https://scrimba.com/) about Fluture and realized it's a pain including all of Fluture's dependencies and wiring them together. This PR adds a step to the `release` script which builds and commits a bundled version with all dependencies included, to be consumed through something like RawGit.

I committed a preview of the build output. Note how the header contains `(dirty)`, this means the build was created outside of the `release` script (detected by the absence of the `VERSION` envvar).

I'm asking your review, @davidchambers, because I'm using the `xyz --script` option for the first time. Can you see anything to improve?

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,13 @@ getPackageName('package.json')
 //> "fluture"
 ```
 
+### From a CDN
+
+Fluture is hosted in full with all of its dependencies at
+https://rawgit.com/fluture-js/Fluture/master/dist/bundle.js.
+
+This script will add `Fluture` to the global scope.
+
 ## Interoperability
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -75,6 +75,13 @@ getPackageName('package.json')
 //> "fluture"
 ```
 
+### From a CDN
+
+Fluture is hosted in full with all of its dependencies at
+`https://rawgit.com/fluture-js/Fluture/master/dist/bundle.js`.
```
davidchambers: Why not make this a link?

Avaq: Ah, yes. Makes sense. :)

PR #150, opened 2017-08-25
Optimize performance of encaseP

### LINKED ISSUE #148: Reason for escapeTick in encaseP
What was the reason for using [this approach](https://github.com/fluture-js/Fluture/blob/master/src/encase-p2.js#L25)?
I use Fluture in an extremely time- and order- sensitive project. Spent a couple of hours to understand that the encaseP resolves by at least 4-12 ms and a few ticks behind the actual resolution of the request, which causes an immediate connection failure.

setTimeout cannot guarantee immediate or exactly next tick execution
https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/setTimeout#Reasons_for_delays_longer_than_specified

Avaq:
**Before**

```text
$ node bench/encase-p.js 
encaseP x 766 ops/sec ±0.68% (57 runs sampled)
encaseP2 x 778 ops/sec ±0.89% (79 runs sampled)
encaseP3 x 775 ops/sec ±0.71% (79 runs sampled)
```

(with an average CPU load of about 2%)

**After**

```text
$ node bench/encase-p.js 
encaseP x 398,000 ops/sec ±0.44% (83 runs sampled)
encaseP2 x 357,165 ops/sec ±4.89% (76 runs sampled)
encaseP3 x 364,562 ops/sec ±1.56% (82 runs sampled)
```

(with an average CPU load of about 80%)

---

Resolves #148

### DIFF
--- a/src/internal/fn.js
+++ b/src/internal/fn.js
@@ -1,5 +1,10 @@
 import Z from 'sanctuary-type-classes';
 import inspectf from 'inspect-f';
+import * as bc from './bc';
+
+const setImmediate = typeof global.setImmediate === 'function'
+  ? global.setImmediate
+  : /* istanbul ignore next: environment-specific */ bc.setImmediate;
 
 export const noop = function noop(){};
 export const moop = function moop(){ return this };
@@ -29,6 +34,4 @@ export const partial3 = (f, a, b, c) => function bound3(d){
   return f(a, b, c, d);
 };
 
-export const escapeTick = f => function imprisoned(x){
-  setTimeout(function escaped(){ f(x) }, 0);
-};
+export const immediately = f => x => setImmediate(f, x);
--- a/bench/encase-p.js
+++ b/bench/encase-p.js
@@ -1,0 +1,32 @@
+'use strict';
+
+const benchmark = require('benchmark');
+const suite = new benchmark.Suite();
+const Future = require('..');
+
+const resolve = x => Promise.resolve(x);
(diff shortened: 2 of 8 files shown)


src/internal/fn.js
```diff
@@ -29,6 +29,9 @@ export const partial3 = (f, a, b, c) => function bound3(d){
...
 
-export const escapeTick = f => function imprisoned(x){
-  setTimeout(function escaped(){ f(x) }, 0);
-};
+export const immediately = typeof global.setImmediate === 'function'
+  ? global.setImmediate
+  : /* istanbul ignore next: environment-specific */
+    (f, x) => setTimeout(f, 0, x);
+
+export const escapeTick = f => function imprisoned(x){ immediately(f, x) };
```
Avaq: Oh - `escapeTick` is just curried `setImmediate`.

src/internal/fn.js
```diff
@@ -1,5 +1,10 @@
 import Z from 'sanctuary-type-classes';
 import inspectf from 'inspect-f';
+import * as bc from './bc';
+
+const setImmediate = typeof global.setImmediate === 'function'
+  ? global.setImmediate
+  : /* istanbul ignore next: environment-specific */ bc.setImmediate;
```
davidchambers: Is `environment-specific` significant or simply a comment?

Avaq: Simply a comment.

PR #160, opened 2017-10-06
Update dependencies in package.json

petercunha:
Just a little housekeeping.


package.json
```diff
@@ -50,39 +50,39 @@
...
   "dependencies": {
-    "concurrify": "^1.0.0",
-    "denque": "^1.1.1",
-    "inspect-f": "^1.2.0",
-    "sanctuary-type-classes": "^6.0.0",
-    "sanctuary-type-identifiers": "^2.0.0"
+    "concurrify": "^1.0.1",
+    "denque": "^1.2.2",
+    "inspect-f": "^1.2.1",
+    "sanctuary-type-classes": "^7.0.0",
```
davidchambers: A “major” upgrade deserves its own pull request so we can review the breaking changes and consider their effects on Fluture.

petercunha:
Alright

package.json
Avaq: See #157. Upgrading to 7.0.0 is technically breaking for users of Fluture.

PR #163, opened 2017-10-13
WIP: Fix type of default export

Avaq:
- [x] Fix the type of the default export. We were telling typescript that the default export *is* a type, where it should have been *of that* type.
- [ ] Fix incompatibilities between TypeScripts own IterableIterator interfae, and Fluture's. When using a generator function in `go()`, TypeScript detects this incompatibility. A possible fix might be to have my own (more sctrict) iterable extend from the less strict TypeScript built-in.

/cc @tetsuo

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -330,7 +330,7 @@ declare module 'fluture' {
   }
 
   export const Future: Fluture
-  export default Fluture
+  export default {} as Fluture
 
   export interface Par {
 


tetsuo:
So, ok, this doesn't work:

```typescript
interface X {
  done: boolean
}

interface Y {
  done: true
}

const a: X = { done: false } // this ok
const b: Y = { done: false } as X // this fails
```

Very _philiosopical_

but, there is at least a proposal to fix Generators in here: https://github.com/Microsoft/TypeScript/issues/2983 Not sure if there's much we can do about it atm, we can type `done` as `{ done: boolean | false }` and `{ done: boolean | true }` in the return result and add a comment referencing that issue.  This will at least throw some information when you hover something 😉

PR #164, opened 2017-10-14
Fix TypeScript generator typings

tetsuo:


index.d.ts
```diff
@@ -17,12 +17,12 @@ declare module 'fluture' {
   }
 
   export interface Next<T> {
-    done: false
+    done: boolean | false
```
Avaq: Is the issue also fixed if you do `false as false`?

tetsuo: `false as false` is syntax error :\ Setting this to just `boolean`

PR #146, opened 2017-08-22
Update rollup to version 0.48.2

Avaq:
Closes #145

PR #166, opened 2017-10-22
Convert source to es5

### LINKED ISSUE #162: Reconsider distribution strategy
Based on [this comment](https://gitter.im/fluture-js/Fluture?at=59a7dc8fbc46472974df45a9) and the conversation preceding it.

Currently, the *module* build of Fluture is only provided as es6. This is a problem for front-end users, who prefer not having to transpile any of their dependencies down to es5. I should distribute a version of Fluture that caters to this group of users.

Furthermore, a new player, [`std/esm`](https://github.com/standard-things/esm) offers an opportunity for shipping the es6 module version directly to NodeJS users without the need to compile to commonjs.

### LINKED ISSUE #165: Failed to minify error on create react app
```
Creating an optimized production build...
Failed to compile.

Failed to minify the code from this file: 

 	./node_modules/fluture/src/internal/throw.js:5 

Read more here: http://bit.ly/2tRViJ9
```

Any ideas on how to fix this?

Avaq:
Gets us the following advantages:

* Better stack traces in unit tests (no need to map them back to the original source)
* Better code coverage reporting (no need to map it through complicated source maps)
* Fewer dependencies (no more compilation dependencies)
* More function names (using named functions everywhere instead of arrow functions)
* No more custom reporter, so upgrade to nyc
* We can run unit tests in all versions of NodeJS
* Faster unit test run time (no compilation step before running tests)
* Faster cjs build (no more need to transpile es6)
* Ready for use in frontend builds

Closes #146 because we don't need the multi-entry rollup plugin anymore.
Closes #162 because this PR is the result of this reconsideration.
Closes #165 because all source is now es5, so usable in front-end builds.


davidchambers:
I'm happy to see this. I believe *libraries* should be written in ES5; *applications* can use whichever build pipeline its developers are comfortable managing.

What about tree shaking? It seems this is finally possible for Ramda (thanks to ES6 modules, I believe). Tree shaking does not seem so important for Fluture, of course.

Avaq:
> What about tree shaking? -- @davidchambers 

You'll notice I'm still using `import` and `export` statements within my es5 code. This will allow tree-shaking tools to do their work.

My strategy is as follows:

* These changes are the first step of a two-step process. The second step will be a breaking change I'll make once NodeJS supports mjs (expected April 2018).
* Currently, though the code uses es5, the module format is esm. For users who rely on cjs I release a cjs build made by Rollup alongside the esm source. In `package.json`, the `main` field points to the cjs build, and the `module` field points to the esm source.
* After Node esm support, I will change all file extensions to `.mjs` (assuming that tree-shaking tools will recognise this extension by then), and point `main` (as well as `module`) to the esm source. At this point I will remove the cjs build from my deployment. Users who *still* rely on cjs will have to run Fluture with the [`@std/esm`](https://github.com/standard-things/esm) runtime.

PR #176, opened 2017-11-01
Fixes: index.d.ts(333,18): error TS2714

mmmveggies:
#164 introduced a bug at https://github.com/fluture-js/Fluture/pull/164/commits/77f102a7d3e6a666ef88447e4e209b13b1b6178e by exporting a default type that was an object.

This reverts it to how it was originally https://github.com/fluture-js/Fluture/commit/48d3a6e0b9fa0058cfea1601ead26e030412311e#diff-b52768974e6bc0faccb7d4b75b162c99R315 

Fixes this:
```
index.d.ts(333,18): error TS2714: The expression of an export assignment must be an identifier or qualified name in an ambient context.
```

This is a compilation problem I get using `fluture` as a dependency, and also adds a test to make sure that's not going to happen again..!

These should result in the same type:
```ts
import Future from 'fluture'
import {Future} from 'fluture'
```

@Avaq why was this changed?

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -330,7 +330,7 @@ declare module 'fluture' {
   }
 
   export var Future: Fluture
-  export default {} as Fluture
+  export default Future
 
   export interface Par {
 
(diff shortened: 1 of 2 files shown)


Avaq:
> why was this changed?

It used to say `export default Fluture`, this was also incorrect.
At the time of fixing it, I missed that it should simply be changed to `Future`, and I tried to work around the issue like that.

This is very nice. Thank you!

Avaq:
^ Released as `7.2.2` :)

Issue #181, opened 2017-11-14
Behaviour of Parallel$cancel


cortopy:
For what I can gather from the [parallel module](https://github.com/fluture-js/Fluture/blob/master/src/parallel.js), the `Parallel$cancel`method would cancel all the Futures which haven't been resolved or rejected yet.

Is this correct?

If so, I was wondering if you would accept a commit to the main docs to say something about this. I'm just using Fluture for the first time and this could be an unnoticed killer feature.

Avaq:
> the `Parallel$cancel`method [cancels] all the Futures which haven't been resolved or rejected yet. Is this correct?

This is correct. Anytime Fluture notices that the result of a process is no longer needed, it is automatically cancelled. Some examples:

* All currently running Futures in `Future.parallel` when one rejects, as you found :) 
* The other in a `Future.race` when the one resolves
* The other in a `race` or `both` when the one rejects

> I was wondering if you would accept a commit to the main docs to say something about this.

Sure thing!

cortopy:
Comments submitted to docs via https://github.com/fluture-js/Fluture/pull/185

PR #185, opened 2017-11-20
mention cancellation logic in parallelism docs

cortopy:
Following https://github.com/fluture-js/Fluture/issues/181, this just adds a few lines to parallelism section in docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -1129,6 +1129,8 @@ first([
 //! "nope"
 ```
 
+When one Future settles, the other gets cancelled automatically.
+
 #### both
 
 <details><summary><code>both :: Future a b -> Future a c -> Future a (Pair b c)</code></summary>
@@ -1150,6 +1152,8 @@ var b = Future.of('b');
 Future.both(a, b).fork(console.error, console.log);
 //> ['a', 'b']
 ```
+
+When one Future rejects, the other gets cancelled automatically.
 
 #### parallel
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1129,6 +1129,8 @@ first([
 //! "nope"
 ```
 
+When one Future resolves, the other gets cancelled automatically.
```
Avaq: Shall we use the term *settles* (instead of *resolves*) to mean *to resolve __or__ reject*?

README.md
```diff
@@ -1151,6 +1153,8 @@ Future.both(a, b).fork(console.error, console.log);
 //> ['a', 'b']
 ```
 
+When one Future resolves, the other gets cancelled automatically.
```
Avaq: This should be *rejects*, rather than *resolves*. When one of the Futures resolves, we wait for the other one to do too, because we are actually awaiting the result of both. :)

cortopy:
Changes done

README.md
```diff
@@ -1129,6 +1129,8 @@ first([
 //! "nope"
 ```
 
+When one Future rejects, the other gets cancelled automatically.
```
Avaq: This one should say *settles*. The `both` one now correctly states *rejects*. When cancellation happens is not the same between the two.

cortopy: Fixed

Avaq:
Thank you! This looks good now. Would you mind squashing the commits and using an uppercase letter in your commit message?

```sh
git fetch
git rebase --interactive origin/master
# Change the first word to "reword" and the rest to "fixup".
# Correct the casing in the commit message, save and exit.
git push origin docs-parallelism-cancel --force
```

Avaq:
Thank you!

PR #192, opened 2017-12-12
Clarify that the rejection reason is also cached

amarshall:
Previously implied that a rejected Future would not be cached, which is untrue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1430,9 +1430,9 @@ cache :: Future a b -> Future a b
 
 </details>
 
-Returns a Future which caches the resolution value of the given Future so that
-whenever it's forked, it can load the value from cache rather than reexecuting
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1430,7 +1430,7 @@ cache :: Future a b -> Future a b
 
 </details>
 
-Returns a Future which caches the resolution value of the given Future so that
+Returns a Future which caches the resolution value or rejection reason of the given Future so that
```
Avaq: Please correct the wrapping on this line, and any following lines which my get affected, such that each line does not exceed the 79 character limit. Sorry for the nitpick.

Avaq:
Would you like to make these small changes to your PR, or do you prefer I do it?

PR #205, opened 2018-01-16
Avoid using "global" explicitly

### LINKED ISSUE #204: Issue when using bundle.js where global throws error for setImmediate
I am trying to use Fluture in Codepen and the CDN bundle.js throws the following error:

`Reference Error: global is not defined`

The instance where global is being called is on the setImmediate function here: https://github.com/fluture-js/Fluture/blob/19febe13038ea5c9d6f2cc499d7b9052a334e611/dist/bundle.js#L2499

I was able to fix this error by wrapping an IIFE that passes the global object like this:

```
var setImmediate = (function(global) {
  return typeof global.setImmediate === 'function'
  ? global.setImmediate
  : /* istanbul ignore next: environment-specific */ setImmediate$1;
})(/*istanbul ignore next*/(commonjsGlobal || window || commonjsGlobal));
```

However, I didn't know if this was how it should be approached in the source file here: 
https://github.com/fluture-js/Fluture/blob/19febe13038ea5c9d6f2cc499d7b9052a334e611/src/internal/fn.js#L5

Avaq:
Fixes #204

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,19 +1,14 @@
 {
+  "root": true,
   "extends": ["warp"],
   "plugins": ["markdown"],
-  "env": {
-    "browser": true
-  },
   "parserOptions": {
     "ecmaVersion": 5,
     "sourceType": "module"
   },
   "globals": {
-    "module": false,
-    "require": false,
-    "global": false,
-    "Future": false,
-    "S": false,
+    "setTimeout": false,
+    "clearTimeout": false,
     "Promise": false
   },
   "rules": {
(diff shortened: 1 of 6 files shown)


Avaq:
I tested it by compiling `dist/bundle.js`, including it in a web-page and constructing a Future. Should that be sufficient @wking-io?

wking-io:
That should do the trick! Thank you for the quick fix and reply!

wking-io:
@Avaq when will the CDN be updated with this version?

Avaq:
I *just* published 8.0.1, which updated the CDN.

PR #220, opened 2018-02-11
Rewrite compliance tests based on Fantasy Laws

Avaq:
The tests should pass once Fantasy Laws is patched with https://github.com/fantasyland/fantasy-laws/pull/4.

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,6 +74,7 @@
     "eslint-config-warp": "^2.0.0",
     "eslint-plugin-markdown": "^1.0.0-beta.7",
     "fantasy-land": "^3.0.0",
+    "fantasy-laws": "^1.0.1",
     "fantasy-states": "^0.2.1",
     "jsverify": "^0.8.3",
     "mocha": "^5.0.2",
--- a/test/7.compliance.test.js
+++ b/test/7.compliance.test.js
@@ -1,124 +1,140 @@
 import {expect} from 'chai';
-import FL from 'fantasy-land';
 import Future from '../index.mjs.js';
-import * as U from './util';
-import jsc from 'jsverify';
+import {assertEqual as eq, I, B, noop, STACKSIZE, error, bang} from './util';
+import {Functor, Bifunctor, Apply, Applicative, Chain, ChainRec, Monad} from 'fantasy-laws';
+import Z from 'sanctuary-type-classes';
+import {
+  bool,
+  constant as _k,
+  falsy,
+  fn,
+  letrec,
+  nat,
+  number,
+  oneof,
+  property,
+  string
+} from 'jsverify';
 
-describe('Compliance', function(){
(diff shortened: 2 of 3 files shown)


test/util.js
```diff
@@ -37,8 +37,14 @@ export var assertEqual = function(a, b){
   b.fork(function(x){bstate = 1; bval = x}, function(x){bstate = 2; bval = x});
   if(astate === 0){ throw new Error('First Future passed to assertEqual did not resolve instantly') }
   if(bstate === 0){ throw new Error('Second Future passed to assertEqual did not resolve instantly') }
+  if(isFuture(aval) && isFuture(bval)) return assertEqual(aval, bval);
```
Avaq: This line is the significant change in this file. It allows for "deep" equality of nested Futures. The only limitation being (of course) that the Futures be synchronous.

test/7.compliance.test.js
```diff
@@ -1,124 +1,140 @@
...
+  var getRight = function(m){
+    var x;
+    m.fork(function(){ throw error }, function(y){ x = y });
+    return x;
+  };
+  return oneof(
+    larb.smap(Future.reject, getLeft, toStr),
+    rarb.smap(Future.of, getRight, toStr),
+    larb.smap(function(x){ return Future.reject(x).map(I) }, getLeft, toStr),
+    rarb.smap(function(x){ return Future.of(x).map(I) }, getLeft, toStr)
```
Avaq: The reason for the latter two Arbitraries in this union is to greatly increase the code coverage made by these tests, because *just* `reject` or `of`-based Futures have optimised implementations of many of the algebras. Mapping over them once with `I` removes this layer of optimisation and falls back to the default implementations.

test/7.compliance.test.js
```diff
@@ -1,124 +1,140 @@
...
+  );
+};
+
+//  anyFuture :: Arbitrary (Future Any Any)
+var {anyFuture} = letrec(function(tie){
```
Avaq: This arbitrary is *by far* the ideal one to use in every test, as it allows tests to assert important properties, such as higher order monads, which would otherwise not be asserted.

Unfortunately, I am unable to use it in most of the tests, as there is no way for me to associated arbitrary data with arbitrary functions that operate on said data. If you'll look into the code below, you'll find that functions are always passed as constants.

I have an idea on how to fix this, and possibly make a bug improvement to fantasy-laws in the process, but for the the scope of the solution is too big to be a concern of this PR.

package.json
```diff
@@ -73,6 +73,7 @@
     "eslint-config-warp": "^2.0.0",
     "eslint-plugin-markdown": "^1.0.0-beta.7",
     "fantasy-land": "^3.0.0",
+    "fantasy-laws": "^1.0.0",
```
davidchambers: This can now be bumped. :)

davidchambers:
:tada:

PR #221, opened 2018-02-12
Use a union for the rejection reason of chain (ts typings)

antogyn:
Right now this scenario fails to compile in typescript:

```typescript
function getA(): Future<AError, A> {}
function getB(a: A): Future<BError, B> {}

getA().chain(getB); // error: Type 'BError' is not assignable to type 'AError'.
```

With the changes in the PR the rejection reason of the future returned by `chain` is the union of the rejection reasons of the base future and the future returned by the mapping function:

```typescript
getA().chain(getB); // type: Future<AError | BError, B> {}
```

The solution without this change is to declare a Future of type `Future<AError | BError, A>` beforehand, which defeats the purpose of type inference.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -67,6 +67,7 @@ declare module 'fluture' {
 
     /** Use the resolution value in this Future to create the next Future. See https://github.com/fluture-js/Fluture#chain */
     chain<RB>(mapper: (value: R) => Future<L, RB>): Future<L, RB>
+    chain<LB, RB>(mapper: (value: R) => Future<LB, RB>): Future<L | LB, RB>
 
     /** Use the rejection reason in this Future to create the next Future. See https://github.com/fluture-js/Fluture#chainrej */
     chainRej<LB>(mapper: (reason: L) => Future<LB, R>): Future<LB, R>
@@ -157,6 +158,8 @@ declare module 'fluture' {
   /** Create a Future using the resolution value of the given Future. See https://github.com/fluture-js/Fluture#chain */
   export function chain<L, RA, RB>(mapper: (value: RA) => Future<L, RB>, source: Future<L, RA>): Future<L, RB>
   export function chain<L, RA, RB>(mapper: (value: RA) => Future<L, RB>): (source: Future<L, RA>) => Future<L, RB>
+  export function chain<LA, LB, RA, RB>(mapper: (value: RA) => Future<LB, RB>, source: Future<LA, RA>): Future<LA | LB, RB>
+  export function chain<LA, LB, RA, RB>(mapper: (value: RA) => Future<LB, RB>): (source: Future<LA, RA>) => Future<LA | LB, RB>
 
   /** Create a Future using the rejection reason of the given Future. See https://github.com/fluture-js/Fluture#chain */
   export function chainRej<LA, LB, R>(mapper: (reason: LA) => Future<LA, R>, source: Future<LA, R>): Future<LB, R>


Avaq:
Hi @antogyn, thank you for your contribution!

However, in many cases, `Type 'BError' is not assignable to type 'AError'` is actually useful information for a developer. I think mixing types *should* require explicit work (such as typing `Future<AError | BError, A>`).

But I may be missing something. Could you show me example values for type `AError` and `BError`?

antogyn:
This change only makes type inference work, but if you use explicit types an error is still returned, e.g.:

```typescript
const b = getA().chain(getB); // type: Future<AError | BError, B> {}

// however, this won't compile (the because the inferred return type is Future<AError | BError, B>) :
function getBfromA(): Future<AError, B> {
  return getA().chain(getB); // Type 'Future<AError | BError, B>' is not assignable to type 'Future<AError, B>'.
}

// you would still need to do something like this:
function getBfromA(): Future<AError | BError, B>  {
  return getA().chain(getB); // works with the change, but does not compile right now: Argument of type '(a: A) => Future<BError, B>' is not assignable to parameter of type '(value: A) => Future<AError, B>'
}

// right now, you need to override the inferred type and do this:
function getBfromA(): Future<AError | BError, B>  {
  return (<Future<AError | BError, A>> getA()).chain(getB);
}
```

Avaq:
Okay. You are on the verge of convincing me. I find the `function getBfromA(): Future<AError, B>`-example quite compelling.

The only thought stopping me is the idea that if you are trying to implicitly create an unprincipled union type, you are *most likely* doing something wrong. This is why I asked you for example values - not to test the code, but to assert the validity of your use case.

So in my view, doing the following is wrong:

```js
function getA(): Future<String, A> {}
function getB(a: A): Future<Error, B> {}

getA().chain(getB); // error: Type 'String' is not assignable to type 'Error'.
```

And generating a union type on the fly is not a fix, it just puts you back in an unsafe world.

Instead, it should be fixed like this:

```js
function getA(): Future<String, A> {}
function getB(a: A): Future<Error, B> {}

getA().mapRej(s => new Error(s)).chain(getB);
```

But I'm not a TypeScript user, so I'm probably not taking something into consideration, which is why I was interested in your specific use-case.

---

**Some time later** I did some looking into this, and it turns out it's one of those polarising subjects in programming. There are those (myself included) who feel Errors should be type safe, because an error handling function that doesn't know its input type seems like a bad idea to me.

But then there are those who have good arguments for the opposing view, like the Swift language team. A colleague of mine recently posted this exact question to the Swift community: https://stackoverflow.com/questions/40718542/why-is-throws-not-type-safe-in-swift

---

In conclusion I will take some extra time mauling this over, and discuss it with other users of Fluture, before making my decision. If we **do** take this route, I believe there are *a lot* more functions than just `chain` to adjust.

davidchambers:
> I will take some extra time mauling this over

I hope you meant to say you will mull it over, Aldwin. ;)

Avaq:
Haha, oops. Updated.

antogyn:
Using union types is actually type safe, you cannot assign a value of type `A | B` to a variable of type `A` in typescript; you first have to determine what the actual type is by differentiating between the types in the union, and in this case make sure that it is of type `A`. The easiest and most idiomatic way to implement this is by using [discriminated unions](https://basarat.gitbooks.io/typescript/docs/types/discriminated-unions.html). This makes it different from the problem you linked in Swift, which is indeed not type safe.

Making type inference "smarter" is actually safer since you don't have to override the inferred type.

In my use case, I have several functions that return Futures which have their own rejection reason, and I want to compose those functions and defer error handling later.

It looks a little like this (except I would generally put the composition in its own function):

```typescript
function validateInputUser(input: InputUser): Future<InputUserValidationError, ValidatedInputUser> {
   ...
}
function fetchUser(validatedInputUser: ValidatedInputUser): Future<IOError, User> {
   ...
}

function handleUserRoute(input: InputUser) {
  return validateInputUser(input)
    .chain(fetchUser)
    .fork(
      (err) => { // err is inferred to be of type `InputUserValidationError | IOError`
        // here we can discriminate between errors and handle them, since err is correctly typed
      },
      (user) => {
         // ...
      }
    )
}

```

Avaq:
Often if I find myself struggling with the hard questions in life, I turn to Haskell and see what it did before me.

```haskell
foo :: Bool -> Either String Bool; foo x = if x then Left "foo" else Right x
bar :: Bool -> Either Bool String; bar x = if x then Left x else Right "bar"

foo True >>= bar
```
```text
error:
    • Couldn't match type ‘Bool’ with ‘[Char]’
      Expected type: Bool -> Either String String
        Actual type: Bool -> Either Bool String
    • In the second argument of ‘(>>=)’, namely ‘bar’
      In the expression: foo True >>= bar
      In an equation for ‘it’: it = foo True >>= bar
```

So it's saying that I cannot use my `bar` function in `chain` (`>>=`), because the Boolean in the Error branch it would produce does not match with the String type it inferred from the use of `foo`.

---

I'm still very much conflicted about this.

On the one hand, I totally see where you are coming from. It's impractical to have a pattern we are used to using so commonly, not work when we enable a static type checker, which has *built in support* for said pattern!

On the other hand, I feel like going down this path is a violation of my own principles (but only *just*). These principles dictate that one *shouldn't* create arbitrary union types, and they are a code smell. And I feel that Fluture (and by extension, the TypeScript typing for Fluture) should guide users to eliminate code smells.

May I ask how exactly you `discriminate between errors and handle them` in this particular case? (My apologies for being such a pain)

Avaq:
Right. I thought about it, and I'm ready to accept this change - I think it's a reasonable trade-off between principle and practicality. Where in principle I still oppose arbitrarily mixing types, but in practice JavaScript allows this anyway, and with TypeScript now forcing you to exhaustively pluck apart your mixed types it seems like a decent thing to allow.

Though I'd prefer not merging this before having gone through all the functions, and adjusted the types to follow this new ideal.

If you scan through the type definitions, you'll find that a lot of functions that combine two Futures use a single type for the rejection branch, and two for the resolution branches. They all need to be changed to use union types instead.

Is this something you'd be willing to do @antogyn, or would you prefer I pushed these changes to your branch?

antogyn:
The Haskell example is interesting, I definitely see where you’re coming from too, what Haskell does is more reasonable/strict. Out of curiosity, what is the most idiomatic way to solve this problem in Haskell?

In this particular case, since typescript doesn’t have nominal typing I make sure that my error types are properly tagged so discriminating is easy.

```typescript
type InputUserValidationError = { kind: 'InputUserValidationError', message: string };
type IOError = { kind: 'IOError', code: number };

// ...

function handleUserRoute(input: InputUser) {
  return validateInputUser(input)
    .chain(createUser)
    .fork(
      (err) => {
        switch (err.kind) {
          case 'InputUserValidationError': {
            // inferred type = InputUserValidationError
            console.log(err.message);
            break;
          }
          case 'IOError': {
            // inferred type = IOError
            console.log(err.code);
            break;
          }
        }
      },
      (user) => {
         // ...
      }
    )
}
```

Glad to see you're ok with this change :).

I don't mind changing other types, I think the following functions would need a change too (did I miss any?) :
- mapRej
- ap
- and
- or
- race
- both
- parallel
- hook
- finally

They all seem trivial except `parallel` since there can be any number of futures; to solve this, generally there is one signature for each possible arity (or length of the array in this case) up to a reasonable/practical number, e.g. the typings for `Promise.all` look like this :

```typescript
// etc, up to 10
all<T1, T2, T3>(values: [T1 | PromiseLike<T1>, T2 | PromiseLike<T2>, T3 | PromiseLike<T3>]): Promise<[T1, T2, T3]>;/
all<T1, T2>(values: [T1 | PromiseLike<T1>, T2 | PromiseLike<T2>]): Promise<[T1, T2]>;
all<T>(values: (T | PromiseLike<T>)[]): Promise<T[]>;
```

which is quite ugly but also quite common, so I don't know if you're ok with this.

Avaq:
`parallel` is the only one I don't want to change. Using it on a mixed Array is simply using it incorrectly. There is `both` for that purpose. See https://github.com/fluture-js/Fluture/pull/184

Avaq:
Making this change would mean that I move away from my previous strategy of closely approximating the Haskell type signatures through TypeScript, towards simply stating in types what the JavaScript implementation semantics are doing.

The signature for `chain` looks like this:

```js
Chain m => (a -> m b) -> m a -> m b
```

Specialised in Future, we substitute `m` for `Future x`, where `x` is the error branch. Is is through this monadic interface that it is not possible to change the error type.

The only reason I'm willing to accept this change, is because there is a group of people (TypeScript users) who do not care about this kind of type consistency. Since I'm providing a service to this particular group, I think I shouldn't force ideals from a different world onto them.

That being said, the line has to be drawn somewhere, or TypeScript becomes useless. If we were to *truly* follow JavaScript semantics, then we would have to overload `+` with all the weird cases, like `{} + [] === 0`.

Currently, I'm willing to draw the line between the resolution and the rejection branch. I don't think it's cool to allow mixed arrays in the resolution branch though using `Future.parallel`. This is simply wrong (like `{} + [] === 0`).

Avaq:
> what is the most idiomatic way to solve this problem in Haskell?

I'll get back to that (if nobody else does) in a day. :)

antogyn:
> Making this change would mean that I move away from my previous strategy of closely approximating the Haskell type signatures through TypeScript, towards simply stating in types what the JavaScript implementation semantics are doing.

> Specialised in Future, we substitute m for Future x, where x is the error branch. Is is through this monadic interface that it is not possible to change the error type.

That makes perfect sense. This change definitely goes against your initial strategy. Thank you for your detailed answers.

> parallel is the only one I don't want to change. Using it on a mixed Array is simply using it incorrectly. There is both for that purpose. See #184 

I see, `parallel` is meant to operate on arrays and not on tuples. Having `all*` methods as you suggested in the other PR would be a good compromise, but using `both` several times is simple enough for me.

Avaq:
> I'll get back to that

I'm by no means a Haskell expert, so I can't say for sure what the *most* idiomatic way is, but I do find that many of my own programming idioms align with those found in the Haskell community, and I can give you three answers.

### 1. Using a strict union type

In Haskell, two types cannot just be smashed together to create a union type. Instead, union types are created by exhaustively listing all the possible constructors. You cannot use already existing type constructors as part of another type, but you can *wrap* existing types. So for example:

```hs
-- "Bool" is the type, True and False are its constructors
data Bool = True | False

-- "Error" is the type, InputUserValidationError and IOError are its constructors
data Error = InputUserValidationError String | IOError Number
```

In this example, `InputUserValidationError` and `IOError` are both constructors for the *same type*. This means that having this in a rejection branch would give us a consistently typed Future as: `Future Error a`. In Haskell you can then use pattern matching to create total functions that operate on every possible case.

To use this kind of approach in JavaScript, you can use one of many union/sum/adt libraries, such as [daggy](https://github.com/fantasyland/daggy). With Daggy you'd use catamorphism to match every possible case. I'll show you. It's going to be a bit involved though, but bear with me: the other two approaches are simpler. The solution to your problem would look somewhat like this:

```js
import {taggedSum} from 'daggy'

const Error = taggedSum('Error', {
  InputUserValidationError: ['message'],
  IOError: ['code']
});

function validateInputUser(input: InputUser): Future<Error, ValidatedInputUser> {
   ...
}
function createUser(validatedInputUser: ValidatedInputUser): Future<Error, User> {
   ...
}

function handleUserRoute(input: InputUser) {
  return validateInputUser(input)
  .chain(createUser)
  .fork((err) => {
    // Catamorphism is like switch/case on type constructors
    // The return value of `cata` is the unified String type,
    // which we can then console.log
    console.log(err.cata({
      InputUserValidationError: message => `Invalid input: ${message}`,
      IOError: code => `I/O error ${String(code)} occurred`
    }))
  }, (user) => {
     // ...
  })
}
```

### 2. Use a simpler type suited for the purpose

Looking at your code, we can acknowledge the following: even though it *seems* you are doing different things for different error types, you're *really* just doing the same thing in both branches: `console.log`. In the `IOError` case you even implicitly coerce the Number to a String in doing so. What if our Error were just a String. In Haskell you'd make a type alias:

```hs
type Error = String
```

In this case, the solution to your problem looks much like what I proposed earlier:

```js
function validateInputUser(input: InputUser): Future<InputUserValidationError, ValidatedInputUser> {
   ...
}
function createUser(validatedInputUser: ValidatedInputUser): Future<IOError, User> {
   ...
}

function handleUserRoute(input: InputUser) {
  return validateInputUser(input)
  .mapRej(({message}) => `Invalid input: ${message}`)
  .map(createUser)
  // I can use chain after converting the IOError to a matching Error type
  .chain(Future.mapRej(({code}) => `I/O error ${String(code)} occurred`))
  // My rejection continuation ends up being nice and simple
  .fork(console.log, (user) => {
    //...
  })
}
```

### 3. Don't smash the types together, and retain choice

Let's say you just used `console.log` as an example, and actually you do *drastically* different things to InputUserValidationError and IOError. When you think about it, your union type only exists to accommodate this *choice*. The thing is, in functional programming we have a special type to represent choice: the Either type. And Future is a superset of the Either type, so we can just use that.

Instead of writing our `switch`/`case` over the two separate types that we smashed together, we can just `fork` twice after *not* having flat-mapped the two inconsistent types:

```js
function validateInputUser(input: InputUser): Future<InputUserValidationError, ValidatedInputUser> {
   ...
}
function createUser(validatedInputUser: ValidatedInputUser): Future<IOError, User> {
   ...
}

function handleUserRoute(input: InputUser) {
  return validateInputUser(input)
  // I'm using map instead of chain
  .map(createUser)
  // I'm forking twice. Semantically this is quite nice, as each possible path is
  // *actually* a fork in our process.
  .fork(handleValidationErrorOneWay, Future.fork(handleIOErrorDrasticallyDifferent, (user) => {
    // ...
  }))
}
```

----

Personally, I usually go with approach `2`. I find that all I do with an error that makes it all the way to `fork`, is log it. Every time. It can just be a String. Sometimes I also like stack traces though, so I use JavaScript's `Error` as my unified rejection type, and my handler always looks like: `e => console.error(e.stack)`.

Avaq:
Along with the other requests to make the TS typings more loosy, I would like to include these changes in a separate typings file, allowing users to opt-in to the loosy variant. I've added a note about these plans in [the roadmap](https://github.com/fluture-js/Fluture/projects/3).

I'm going to close this PR, because it doesn't include typings for similar transformations to `chain`, and because of the different approach I would like to use. Thank you for bringing your perspective to my attention, @antogyn, and my apologies for my somewhat ranty answer in the end.

PR #222, opened 2018-02-12
Prevent running task in parallel after one rejects

Avaq:


davidchambers:
I don't understand the fix, but I like one-line fixes. :)

Avaq:
`cursor = _length` just causes the break condition in the while-loop a few lines below.

PR #224, opened 2018-02-15
Catch exceptions

### LINKED ISSUE #153: Unable to recover from thrown exceptions
Below is an example of the issue.

```js
'use strict';

const instantCache = Future.cache(Future.of('instantVal'));
const slowCache = Future.cache(Future((rej, res) => {setTimeout(() => res('slowVal'), 50)}));
const brokenChainMapper = () => { throw new Error('Broken') };
const noop = () => {}

try {
	instantCache.chain(brokenChainMapper).fork(noop, noop);
} catch(e) {
	console.log(`Caught instant cache error ${e.message}`);
}

try {
	slowCache.chain(brokenChainMapper).fork(noop, noop);
} catch(e) {
	console.log(`Caught slow cache error ${e.message}`);
}
```

Here the `instantCache` version gets caught and logged. While the `slowCache` one crashes the process like an uncaught exception.

My usecase is to cache a DB connection used for web server request handlers. Then using the try...catch as a last resort to output 500 errors, rather than crash the process.

Tested on node 8.5.0 and fluture 7.1.2.

Avaq:
So I finally reached a point where all the original unit tests are passing again. This marks an important milestone in my work towards implementing a solution for #153.

But the suspicion I mentioned having in #196 has been confirmed: These changes appear to affect performance significantly. In particular the creation of Futures has slowed down, in many cases to about half of the speed, in some even to 10% of the original speed. I don't yet understand why this happens.

latest-screenshot

Additional things to do:

- [x] Restore test coverage
- [x] Ensure the recovery branch has a consistent type
- [x] Ensure `fork` is not used in tests asserting `_interpret` - also rename test cases
- [x] Ensure that the `@@type` property is bumped
- [x] Fix the performance issues

/cc @beanow


Avaq:
The performance characteristics actually look quite different in Node v8. I was using v9 before.

```text
┌─────────────────────────────────────────┬────────────────────────────┬────────────────────────────┬────────┬─────────┬───┐                                                                   
│ suite                                   │ Old                        │ New                        │ diff   │ change  │ α │
├─────────────────────────────────────────┼────────────────────────────┼────────────────────────────┼────────┼─────────┼───┤
│ def.construct.Future                    │ 7,998,149 Hz ±0.68% (n 84) │ 7,258,871 Hz ±1.14% (n 83) │ 004.8% │ -009.2% │   │
├─────────────────────────────────────────┼────────────────────────────┼────────────────────────────┼────────┼─────────┼───┤
│ def.construct.of                        │ 7,261,358 Hz ±0.86% (n 84) │ 6,617,226 Hz ±1.37% (n 79) │ 004.6% │ -008.9% │   │
├─────────────────────────────────────────┼────────────────────────────┼────────────────────────────┼────────┼─────────┼───┤
│ def.construct.reject                    │ 7,370,939 Hz ±0.69% (n 82) │ 6,045,733 Hz ±0.76% (n 83) │ 009.9% │ -018.0% │   │
├─────────────────────────────────────────┼────────────────────────────┼────────────────────────────┼────────┼─────────┼───┤
│ def.construct.after                     │ 6,954,759 Hz ±1.51% (n 83) │ 5,605,209 Hz ±0.48% (n 85) │ 010.7% │ -019.4% │ ✗ │
├─────────────────────────────────────────┼────────────────────────────┼────────────────────────────┼────────┼─────────┼───┤
│ def.construct.attempt                   │ 6,064,616 Hz ±2.66% (n 81) │ 4,434,121 Hz ±5.59% (n 64) │ 015.5% │ -026.9% │ ✗ │
├─────────────────────────────────────────┼────────────────────────────┼────────────────────────────┼────────┼─────────┼───┤
│ def.construct.cache                     │ 4,264,237 Hz ±7.22% (n 67) │ 3,729,538 Hz ±4.98% (n 71) │ 006.7% │ -012.5% │   │
├─────────────────────────────────────────┼────────────────────────────┼────────────────────────────┼────────┼─────────┼───┤
│ def.construct.chainRec                  │ 5,914,881 Hz ±2.39% (n 78) │ 5,387,866 Hz ±1.54% (n 82) │ 004.7% │ -008.9% │   │
├─────────────────────────────────────────┼────────────────────────────┼────────────────────────────┼────────┼─────────┼───┤
│ def.construct.encase                    │ 5,663,542 Hz ±1.48% (n 81) │ 4,448,120 Hz ±2.83% (n 74) │ 012.0% │ -021.5% │ ✗ │
├─────────────────────────────────────────┼────────────────────────────┼────────────────────────────┼────────┼─────────┼───┤
│ def.construct.encaseN                   │ 4,734,732 Hz ±6.55% (n 65) │ 4,182,550 Hz ±4.26% (n 72) │ 006.2% │ -011.7% │   │
├─────────────────────────────────────────┼────────────────────────────┼────────────────────────────┼────────┼─────────┼───┤
│ def.construct.encaseP                   │ 5,494,980 Hz ±2.93% (n 78) │ 5,058,328 Hz ±0.81% (n 79) │ 004.1% │ -007.9% │   │
├─────────────────────────────────────────┼────────────────────────────┼────────────────────────────┼────────┼─────────┼───┤
│ def.construct.go                        │ 6,073,758 Hz ±1.91% (n 76) │ 5,529,125 Hz ±1.60% (n 80) │ 004.7% │ -009.0% │   │
├─────────────────────────────────────────┼────────────────────────────┼────────────────────────────┼────────┼─────────┼───┤
│ def.construct.hook                      │ 4,436,183 Hz ±2.24% (n 75) │ 3,745,082 Hz ±1.48% (n 84) │ 008.4% │ -015.6% │   │
├─────────────────────────────────────────┼────────────────────────────┼────────────────────────────┼────────┼─────────┼───┤
│ def.construct.node                      │ 6,075,830 Hz ±1.22% (n 83) │ 5,242,111 Hz ±1.09% (n 83) │ 007.4% │ -013.7% │   │
├─────────────────────────────────────────┼────────────────────────────┼────────────────────────────┼────────┼─────────┼───┤
│ def.construct.parallel                  │ 4,740,475 Hz ±0.70% (n 80) │ 3,109,314 Hz ±3.18% (n 77) │ 020.8% │ -034.4% │ ✗ │
├─────────────────────────────────────────┼────────────────────────────┼────────────────────────────┼────────┼─────────┼───┤
│ def.transform.ap                        │ 130,686 Hz ±3.71% (n 75)   │ 118,276 Hz ±4.62% (n 69)   │ 005.0% │ -009.5% │   │
├─────────────────────────────────────────┼────────────────────────────┼────────────────────────────┼────────┼─────────┼───┤
│ def.transform.map                       │ 361,595 Hz ±5.57% (n 69)   │ 433,094 Hz ±2.07% (n 80)   │ 009.0% │ +019.8% │   │
├─────────────────────────────────────────┼────────────────────────────┼────────────────────────────┼────────┼─────────┼───┤
│ def.transform.chain                     │ 174,833 Hz ±2.76% (n 81)   │ 155,470 Hz ±4.46% (n 71)   │ 005.9% │ -011.1% │   │
├─────────────────────────────────────────┼────────────────────────────┼────────────────────────────┼────────┼─────────┼───┤
│ run.construct.parallel.empty            │ 5,790,409 Hz ±2.71% (n 81) │ 4,172,110 Hz ±3.67% (n 74) │ 016.2% │ -027.9% │ ✗ │
├─────────────────────────────────────────┼────────────────────────────┼────────────────────────────┼────────┼─────────┼───┤
│ run.construct.parallel.small.sequential │ 366,858 Hz ±3.93% (n 74)   │ 375,237 Hz ±2.77% (n 74)   │ 001.1% │ +002.3% │   │
├─────────────────────────────────────────┼────────────────────────────┼────────────────────────────┼────────┼─────────┼───┤
│ run.construct.parallel.small.concurrent │ 333,434 Hz ±4.10% (n 71)   │ 380,925 Hz ±2.69% (n 76)   │ 006.6% │ +014.2% │   │
├─────────────────────────────────────────┼────────────────────────────┼────────────────────────────┼────────┼─────────┼───┤
│ run.construct.parallel.big.sequential   │ 14,344 Hz ±4.75% (n 60)    │ 16,291 Hz ±6.95% (n 70)    │ 006.4% │ +013.6% │   │
├─────────────────────────────────────────┼────────────────────────────┼────────────────────────────┼────────┼─────────┼───┤
│ run.construct.parallel.big.concurrent   │ 19,117 Hz ±2.92% (n 78)    │ 19,891 Hz ±1.49% (n 81)    │ 002.0% │ +004.0% │   │
├─────────────────────────────────────────┼────────────────────────────┼────────────────────────────┼────────┼─────────┼───┤
│ run.transform.sync.ap                   │ 112,592 Hz ±2.48% (n 77)   │ 109,112 Hz ±2.37% (n 78)   │ 001.6% │ -003.1% │   │
├─────────────────────────────────────────┼────────────────────────────┼────────────────────────────┼────────┼─────────┼───┤
│ run.transform.sync.map                  │ 298,304 Hz ±0.92% (n 81)   │ 248,961 Hz ±4.87% (n 70)   │ 009.0% │ -016.5% │   │
├─────────────────────────────────────────┼────────────────────────────┼────────────────────────────┼────────┼─────────┼───┤
│ run.transform.sync.chain.one            │ 99,727 Hz ±9.64% (n 59)    │ 125,796 Hz ±2.92% (n 74)   │ 011.6% │ +026.1% │ ✓ │
├─────────────────────────────────────────┼────────────────────────────┼────────────────────────────┼────────┼─────────┼───┤
│ run.transform.sync.chain.many           │ 149 Hz ±4.68% (n 63)       │ 164 Hz ±3.68% (n 68)       │ 004.7% │ +009.9% │   │
├─────────────────────────────────────────┼────────────────────────────┼────────────────────────────┼────────┼─────────┼───┤
│ run.transform.async.ap                  │ 51,058 Hz ±1.89% (n 80)    │ 47,973 Hz ±3.26% (n 73)    │ 003.1% │ -006.0% │   │
├─────────────────────────────────────────┼────────────────────────────┼────────────────────────────┼────────┼─────────┼───┤
│ run.transform.async.map                 │ 97,264 Hz ±2.62% (n 75)    │ 92,465 Hz ±2.03% (n 75)    │ 002.5% │ -004.9% │   │
├─────────────────────────────────────────┼────────────────────────────┼────────────────────────────┼────────┼─────────┼───┤
│ run.transform.async.chain.one           │ 57,758 Hz ±1.72% (n 75)    │ 55,297 Hz ±2.03% (n 75)    │ 002.2% │ -004.3% │   │
├─────────────────────────────────────────┼────────────────────────────┼────────────────────────────┼────────┼─────────┼───┤
│ run.transform.async.chain.many          │ 100 Hz ±4.14% (n 73)       │ 99 Hz ±2.95% (n 73)        │ 000.4% │ -000.8% │   │
├─────────────────────────────────────────┼────────────────────────────┼────────────────────────────┼────────┼─────────┼───┤
│ run.transform.async.race.fast-vs-slow   │ 59,531 Hz ±11.25% (n 71)   │ 64,229 Hz ±4.42% (n 75)    │ 003.8% │ +007.9% │   │
├─────────────────────────────────────────┼────────────────────────────┼────────────────────────────┼────────┼─────────┼───┤
│ run.transform.async.race.slow-vs-fast   │ 69,577 Hz ±3.11% (n 75)    │ 64,617 Hz ±2.15% (n 74)    │ 003.7% │ -007.1% │   │
├─────────────────────────────────────────┼────────────────────────────┼────────────────────────────┼────────┼─────────┼───┤
│ run.transform.async.race.slow-vs-slow   │ 782 Hz ±1.57% (n 78)       │ 780 Hz ±0.81% (n 77)       │ 000.1% │ -000.2% │   │
└─────────────────────────────────────────┴────────────────────────────┴────────────────────────────┴────────┴─────────┴───┘
```

Avaq:
Given:

```js
const unsafe = Math.random() > 0.5 ? {foo: 1} : null
const noop = () => {}
```

I'm a bit concerned that the following:

```js
Future((rej, res) => {
  const foo = unsafe.foo
  res(foo)
})._interpret(noop, noop, noop)
```

Will not be equal to:

```js
const foo = unsafe.foo
Future((rej, res) => {
  res(foo)
})._interpret(noop, noop, noop)
```

Essentially the exception catching feature removes a level of being able to reason about, and safely refactor, certain code.

Beanow:
@Avaq I would not have considered these to be equivalent either way. Especially when considering potential failures.

Should `unsafe.foo` for example have been a configuration value, I would strongly favour doing related evaluations as early as possible, so bad configuration can crash the program early. Additionally such a future might be generated more often by a factory it'd be wasteful to evaluate it repeatedly (unless `unsafe` is supposed to be mutable).

On the other hand, if this `unsafe.foo` is a value obtained through less predictable means, such as a database or external API, it would be swell to contain the failure and prevent crashes. Even if it would arguably be an implementation error to be fixed in the code, rather than a user error.

index.mjs.js
```diff
@@ -1,16 +1,10 @@
-import {error, invalidArgument} from './src/internal/throw';
-
-if(typeof Object.create !== 'function') error('Please polyfill Object.create to use Fluture');
-if(typeof Object.assign !== 'function') error('Please polyfill Object.assign to use Fluture');
-if(typeof Array.isArray !== 'function') error('Please polyfill Array.isArray to use Fluture');
```
Avaq: This chunk was useless, because `rollup` changes the order of code evaluation in such a way that this code is reached only after some parts have already tried to use these functions. It's also not super important, as it only served to slightly improve the error messages.

index.mjs.js
```diff
@@ -20,7 +14,7 @@ Future.map = map;
 Future.bimap = bimap;
 Future.chain = chain;
 
-var Par = concurrify(Future, never, race, parallelAp);
+var Par = concurrify(Future, never, race, function parallelAp(a, b){ return b._parallelAp(a) });
```
Avaq: I changed `parallelAp` from a static constructor to a standard Action. This saved me writing some extra code.

package.json
```diff
@@ -18,13 +18,16 @@
...
     "release": "npm outdated --long; xyz --edit --repo git@github.com:fluture-js/Fluture.git --tag 'X.Y.Z' --script scripts/distribute --increment",
-    "test": "npm run lint && npm run lint:readme && npm run test:unit && npm run test:types",
-    "test:unit": "npm run build && nyc --include src mocha --require @std/esm --ui bdd --reporter list --check-leaks --full-trace test/**.test.js",
-    "test:coverage": "nyc report --reporter=text-lcov > coverage.lcov && codecov",
+    "test": "npm run lint && npm run lint:readme && npm run test:unit && npm run test:types && npm run test:build",
+    "test:unit": "npm run clean && mocha --require @std/esm --ui bdd --reporter list --full-trace --check-leaks --bail test/**.test.js",
+    "test:build": "npm run clean && npm run build && mocha --require @std/esm --ui bdd --reporter dot --bail test/**.buildtest.js",
+    "test:coverage": "npm run clean && nyc --include src mocha --require @std/esm --ui bdd --reporter dot test/**.test.js || true",
+    "coverage:upload": "nyc report --reporter=text-lcov > coverage.lcov && codecov",
+    "coverage:report": "nyc report --reporter=html",
```
Avaq: These changes facilitate potentially failing tests while testing for coverage -- annoyingly when testing `toString`, sometimes the coverage instrumentation breaks expected output.

Avaq:
I just flipped the libraries I'm comparing, so `Old` shows the results for the new stuff, and `New` for the old stuff, and it turns out that the results are exactly the same.. So we were getting false positives.

```text
┌────────────────────────┬──────────────────────────────┬─────────────────────────────┬────────┬─────────┬───┐                                                                                 
│ suite                  │ Old                          │ New                         │ diff   │ change  │ α │
├────────────────────────┼──────────────────────────────┼─────────────────────────────┼────────┼─────────┼───┤
│ def.construct.Future   │ 277,883,345 Hz ±0.66% (n 90) │ 25,408,997 Hz ±1.12% (n 83) │ 083.2% │ -090.9% │ ✗ │
├────────────────────────┼──────────────────────────────┼─────────────────────────────┼────────┼─────────┼───┤
│ def.construct.of       │ 166,840,514 Hz ±0.38% (n 86) │ 30,217,530 Hz ±1.00% (n 85) │ 069.3% │ -081.9% │ ✗ │
├────────────────────────┼──────────────────────────────┼─────────────────────────────┼────────┼─────────┼───┤
│ def.construct.reject   │ 177,376,212 Hz ±1.21% (n 78) │ 30,463,316 Hz ±2.22% (n 83) │ 070.7% │ -082.8% │ ✗ │
├────────────────────────┼──────────────────────────────┼─────────────────────────────┼────────┼─────────┼───┤
│ def.construct.after    │ 46,805,158 Hz ±0.48% (n 83)  │ 19,378,201 Hz ±0.93% (n 81) │ 041.4% │ -058.6% │ ✗ │
├────────────────────────┼──────────────────────────────┼─────────────────────────────┼────────┼─────────┼───┤
│ def.construct.attempt  │ 26,059,718 Hz ±0.74% (n 86)  │ 16,275,959 Hz ±1.09% (n 79) │ 023.1% │ -037.5% │ ✗ │
├────────────────────────┼──────────────────────────────┼─────────────────────────────┼────────┼─────────┼───┤
│ def.construct.cache    │ 10,537,069 Hz ±1.87% (n 87)  │ 9,691,659 Hz ±0.91% (n 91)  │ 004.2% │ -008.0% │   │
├────────────────────────┼──────────────────────────────┼─────────────────────────────┼────────┼─────────┼───┤
│ def.construct.chainRec │ 18,929,958 Hz ±2.29% (n 83)  │ 11,663,289 Hz ±1.81% (n 75) │ 023.8% │ -038.4% │ ✗ │
├────────────────────────┼──────────────────────────────┼─────────────────────────────┼────────┼─────────┼───┤
│ def.construct.encase   │ 20,333,009 Hz ±1.76% (n 76)  │ 11,697,816 Hz ±0.90% (n 88) │ 027.0% │ -042.5% │ ✗ │
├────────────────────────┼──────────────────────────────┼─────────────────────────────┼────────┼─────────┼───┤
│ def.construct.encaseN  │ 18,895,576 Hz ±1.30% (n 82)  │ 10,578,323 Hz ±0.99% (n 85) │ 028.2% │ -044.0% │ ✗ │
├────────────────────────┼──────────────────────────────┼─────────────────────────────┼────────┼─────────┼───┤
│ def.construct.encaseP  │ 18,758,250 Hz ±0.72% (n 86)  │ 9,836,303 Hz ±0.77% (n 90)  │ 031.2% │ -047.6% │ ✗ │
├────────────────────────┼──────────────────────────────┼─────────────────────────────┼────────┼─────────┼───┤
│ def.construct.go       │ 16,231,686 Hz ±2.15% (n 77)  │ 13,508,766 Hz ±0.58% (n 92) │ 009.2% │ -016.8% │   │
├────────────────────────┼──────────────────────────────┼─────────────────────────────┼────────┼─────────┼───┤
│ def.construct.hook     │ 9,871,054 Hz ±0.64% (n 90)   │ 8,061,882 Hz ±0.69% (n 89)  │ 010.1% │ -018.3% │ ✗ │
├────────────────────────┼──────────────────────────────┼─────────────────────────────┼────────┼─────────┼───┤
│ def.construct.node     │ 27,726,998 Hz ±0.94% (n 80)  │ 12,686,311 Hz ±0.71% (n 85) │ 037.2% │ -054.2% │ ✗ │
├────────────────────────┼──────────────────────────────┼─────────────────────────────┼────────┼─────────┼───┤
│ def.construct.parallel │ 9,710,786 Hz ±1.15% (n 90)   │ 6,811,319 Hz ±0.48% (n 91)  │ 017.5% │ -029.9% │ ✗ │
└────────────────────────┴──────────────────────────────┴─────────────────────────────┴────────┴─────────┴───┘
```

Avaq:
Seeing as the diff is so huge, I would like to highlight some of the important changes and decisions, so that they can be discussed.

1. **Consistently typed recovery branch.** When Fluture catches a value that was thrown, it is turned into a `new Error` no matter the original value. The message of this error will contain any message that could be extracted from the value thrown, plus some contextual information.
  https://github.com/fluture-js/Fluture/blob/85d4eb8cbb3048336701d887fffc31804a3e2f63/src/internal/error.js#L61-L76
  This means that users will not be able to use mixing error types to handle them differently, this is by design - an error in the recovery branch is *always* considered a bug. It also means that users can rely on the recovery branch having a consistent type.

2. **Asynchronously thrown exceptions.** When a Future is forked, it might synchronously call the continuations if the underlying computation happens synchronously. This meant that if an exception would be thrown, some outer `catch` statement would sometimes get the opportunity to catch it. I've removed this opportunity by having `fork` *always* throw the exception in the recovery branch *in the next tick*.
  https://github.com/fluture-js/Fluture/blob/85d4eb8cbb3048336701d887fffc31804a3e2f63/src/internal/fn.js#L43-L45
  https://github.com/fluture-js/Fluture/blob/85d4eb8cbb3048336701d887fffc31804a3e2f63/src/core.js#L129-L134
  This means that if you don't specify a recovery function, a bug will *always* crash the program (or end up in a global exception handler, which is expected to then exit the program).
  It also means that `Future((rej, res) => { otherFuture().fork(rej, compose(res, f)) })` is no longer strictly equivalent to `otherFuture().map(f)` in cases where `otherFuture()` resolves synchronously. To me that sounds like a small price to pay.

3. **Throwing the recovery reason on `.promise()`.** Promises mix bugs with expected failures and in my view that's a flaw. Fluture keeps them separate. But what to do when casting a Future to a Promise? Currently the error is thrown.
  https://github.com/fluture-js/Fluture/blob/85d4eb8cbb3048336701d887fffc31804a3e2f63/src/core.js#L150-L155
  This means that `.promse()` potentially throws (which has always been the case). But since users asked for a Promise, they might actually want the bugs and failures to mix.

rpominov:
> This meant that if an exception would be thrown, some outer catch statement would sometimes get the opportunity to catch it.

It's kinda hard for me to judge because I wouldn't wrap big chunks of computation into a `try...catch`, but maybe for people who do, it's actually desirable to have such kind of exceptions caught. 

Also, I'm guessing that the solution involves a rethrowing an exception. In this case, a debugger won't pause on the original exception, and users won't be able to inspect a frozen program. I, for one, rely a lot on this feature of a debugger.

> This means that .promse() potentially throws (which has always been the case). But since users asked for a Promise, they might actually want the bugs and failures to mix.

A possible solution would be to direct bugs into promises' failure path and direct both errors and normal results into promises' success, using an improvised Either. Basically what fun-task does https://github.com/rpominov/fun-task/blob/master/docs/api-reference.md#tasktopromiseoptions 😅

Avaq:
>> This meant that if an exception would be thrown, some outer catch statement would sometimes get the opportunity to catch it.
>
> It's kinda hard for me to judge because I wouldn't wrap big chunks of computation into a `try...catch`

Generally this kind of thing would happens if there is some underlying abstraction calling your code, such as when the Future is forked as part of an Express middleware.

> A possible solution would be to direct bugs into promises' failure path and direct both errors and normal results into promises' success, using an improvised Either.

I like this idea, but I don't like introducing a competing Either type. Perhaps I can change the signature to:

```hs
promise :: (a -> c) -> (b -> c) -> Future a b -> Promise c
```

So a user would be able to use it as:

```js
const toPromise = Future.promise(S.Left, S.Right);

toPromise(Future.reject('meh')) // Promise<Left('meh')>
```

Avaq:
> I can get very angry at times when a library messes with exceptions somehow

My reasoning for this behaviour was that I wanted to make the behaviour more consistent. Nine out of ten times when you `fork` a Future that ends up throwing, it will throw halfway through the computation in some unexpected tick. The only time you can catch an exception thrown during computation is if it happens to throw in the same tick in which `.fork` was called. In my experience this can lead to surprises. Eliminating the possibility for catching these exceptions altogether seemed to me like it would lead to fewer surprises.

However, I think it's a very compelling point that this is kind of anti-tooling. It's probably easier to track an error that was thrown, caught, and retrown all in the same tick, using various debuggers. I also like the idea that `Future((rej, res) => { otherFuture().fork(rej, compose(res, f)) })` remains equivalent to `otherFuture().map(f)`. This allows for people to write their own transformations using `Future()` and `fork()`, without losing something over the built-in transformations.

So in conclusion, I will experiment with synchronously thrown errors, and likely amend this PR. Thank you for your feedback Roman! :D

Avaq:
> `Future.promise(S.Left, S.Right)`

I'm going to leave this for another PR, leaving the signature (and behaviour) of `.promise()` unchanged for this one.

PR #230, opened 2018-03-05
Use type variables in TypeScript typings instead of never 

tetsuo:
```
F.of(x => y => [x, y])
  .ap(future1)
  .ap(future2)
```

TSC throws a compilation error in here, since `of` always assumes that it doesn't throw a rejection.

Adding a new type variable instead using never should fix the issue, but it might break other things where the error type is not specified explicitly.

@Avaq I added it to `of` and `after`. Anything else that needs to be fixed?

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -87,7 +87,7 @@ declare module 'fluture' {
     finally(cleanup: Future<L, any>): Future<L, R>
 
     /** Fold both branches into the resolution branch. See https://github.com/fluture-js/Fluture#fold */
-    fold<RB>(lmapper: (reason: L) => RB, rmapper: (value: R) => RB): Future<never, RB>
+    fold<L, RB>(lmapper: (reason: L) => RB, rmapper: (value: R) => RB): Future<L, RB>
 
     /** Fork the Future into the two given continuations. See https://github.com/fluture-js/Fluture#fork */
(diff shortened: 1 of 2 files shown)


Avaq:
I believe this approach is the better one in all instances where `never` was used before.

Avaq:
Would you like to make the proposed adjustment to all the other `never` using functions (except `value`), @tetsuo? Or shall I make these changes and append them to your PR?

tetsuo:
Well, I was working on that, but ran into some deep questions at some point :) can we finish it off together some time this week?

tetsuo:
The final build requires `@types/node` typings for some reason, which is installed for Node.js v10 by default. This was causing builds to fail during declaration file testing. It looks like the default node typings require `Map`, `Set` and so on. I included ES6 typings too. We can also try installing one of the earlier versions of `@types/node`. Also updated TypeScript to `2.8.x` just for kicks.

Should we consider adding a `package-lock.json` file by the way?

Avaq:
This looks great! Thank you.

I remain of the opinion that package locks are only relevant in "root" level packages, eg applications. It's a feature to accommodate safer deployments. When installing a package as a dependency, I believe its package lock is ignored. Instead, I use [greenkeeper](https://greenkeeper.io/) to react to broken dependencies.

PR #238, opened 2018-04-06
Improve correctness of Future.hook

### LINKED ISSUE #234: Make dispose in Hook uncancellable
Based on [this Gitter conversation](https://gitter.im/fluture-js/Fluture?at=5aa510a1e4d1c63604e85f03) and the PureScript Aff [documentation of `bracket`](https://github.com/slamdata/purescript-aff#3-bracketing) it seems to me that `Future.hook` would be more correct if it would never cancel the `dispose` Future.

Avaq:
This PR introduces *two* significant changes to how Hook works:

1. The cancellation signal is never sent down to the disposal Future anymore. If cancellation happens, disposal is allowed to complete, but its results are ignored. Closes #234.
2. When an exception occurs during consumption, the disposal Future *still runs*, similarly to what happens when the consumption Future is cancelled. This was not considered while exception recovery was being implemented. Now that exception recovery exists, we should allow resources to be disposed (we no longer assume to process to crash).

I've tried to explain these semantics [in the README](https://github.com/fluture-js/Fluture/tree/avaq/hook-correctness#hook).

@safareli I've asked you for a review because this change affects semantics and is partly inspired by PureScript Aff.

Ping @syaiful6 because you were the person to bring this to my attention.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1347,31 +1347,21 @@ withConnection(
 .fork(console.error, console.log);
 ```
 
-In the case that a hooked Future is *cancelled* after the resource was acquired,
-`dispose` will be executed and immediately cancelled. This means that rejections
-which may happen during this disposal are **silently ignored**. To ensure that
-resources are disposed during cancellation, you might synchronously dispose
-resources in the `cancel` function of the disposal Future:
+When a hooked Future is cancelled while acquiring its resource, nothing else
+will happen. However, once the resource has been acquired, the `dispose`
+Future is guaranteed to run.
 
-<!-- eslint-disable no-unused-vars -->
-```js
-var closeConnection = conn => Future((rej, res) => {
+In the case where the consumption Future settles, the disposal is forked and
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1347,31 +1347,21 @@ withConnection(
...
-      conn.close();
-      res();
-    }else{
-      rej(err);
-    }
-  });
+In the case where the consumption Future settles, the disposal is forked and
+awaited. If the disposal Future rejects or encounters an exception, it will
+replace the state of the consumption Future. Otherwise the rejection reason or
+resolution value of the consumption Future will determine the final outcome.
```
davidchambers: Is “consumption” being used as an adjective here? This doesn't sound right to me.

Avaq: I need to quick way to refer to "the Future returned by the `consume` function".

davidchambers: Fair enough.

Avaq:
> I can't do the review in close future :(

Alright, no worries!

Issue #240, opened 2018-04-16
encased functions have the wrong arity


semmel:
When called with just the promise-returning function argument, the `encaseP*` family of functions always return a function `bound1` of arity 3. 

```javascript
var square = function(x){ return Promise.resolve(x * x); };
var squared = F.encaseP(square);
squared.length // -> 3
```

This prevents the use of the encased functions with other libraries like Ramda which auto-curry variadic user functions. This example code is a stripped down version of my problem:

```javascript
F.of(10).chain(
	R.ifElse(
		R.gt(R.__, 9), 
		squared, 
		F.of
	)
).fork(console.log, console.log);
// -> TypeError: Future#chain expects the function it's given to return a Future.
//  Actual: function (a0, a1)...
// The reason is that R.ifElse returns a curryN function with N being the max. arity
// of the truthy and falsy functions which is 3 here.

// Work-Around: Use proper partial application
var squaredToo = R.partial(F.encaseP, [square]);
squaredToo.length // -> 1
F.of(10).chain(
	R.ifElse(
		R.gt(R.__, 9), 
		squaredToo, 
		F.of
	)
).fork(console.log, console.log); // -> 100
```
While a fix should be straight-forward I ask myself if it is worth it. One could remove currying altogether and
* leave currying the Fluture's api to external libs like Ramda or underscore, and
* prefer explicit partial application of the arguments over auto-currying, since that removes ambiguity from my code.

On a related point I find myself in need for `encaseP0`. Anyone else? Should be simple enough for me to come up with a pull request...

semmel:
> On a related point I find myself in need for encaseP0

Just realized that `encaseP0 = R.compose(R.partial(F.tryP), R.of)`!
```javascript
var fortyTwo = function(){ return Promise.resolve(42); };
var encaseP0 = R.compose(R.partial(F.tryP), R.of);

encaseP0(fortyTwo)()
.fork(console.log, console.log); // -> 42
```

Avaq:
> the `encaseP*` family of functions always return a function `bound1` of arity 3.

Actually it's all Fluture functions that are partially applied. The currying mechanism in Fluture makes compromises to gain better performance. One such compromise makes partially applied functions report the "wrong" arity. I find that the only time this is problematic is when using Ramda's variadic composition functions.

> One could remove currying altogether and leave currying the Fluture's api to external libs like Ramda or underscore.

[...]

PR #250, opened 2018-06-06
Switch to sanctuary-show for string representations

Avaq:
Closes #244


Avaq:
> changing `Future(…)` to `Future (…)`.

I might at a later stage. I'm also considering switching to simple currying (#240), if and when I do, I'll change the function application style along with it.

PR #261, opened 2018-07-19
Add forkCatch - fork with exception recovery

### LINKED ISSUE #153: Unable to recover from thrown exceptions

Avaq:
Closes #153

/cc @beanow

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,6 +171,7 @@ sponsoring the project.
 <details><summary>Consuming/forking Futures</summary>
 
 - [`fork`: Standard way to run a Future and get at its result](#fork)
+- [`forkCatch`: Fork with exception recovery](#forkcatch)
 - [`value`: Shorter variant of `fork` for Futures sure to succeed](#value)
 - [`done`: Nodeback style `fork`](#done)
 - [`promise`: Convert a Future to a Promise](#promise)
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -1017,6 +1018,37 @@ var cancel = fut.fork(console.error, console.log);
...
+
+Generally it's best to let a program crash and restart when an unexpected
+exception occurs. Restarting is the surest way to restore the memory that was
+allocated by the program to an expected state.
+
+By using `forkCatch`, you might be able to keep your program alive longer,
+which can be very beneficial when the program is being used by multiple
+clients. However, you also forego the certainty that your program will be in a
+valid state after this happens. The more isolated the memory consumed by the
+particular computation was, the more certain you will be that recovery is safe.
```
davidchambers: Nice explanation. :)

PR #266, opened 2018-07-26
Rename TypeScript interfaces not to conflict with named values

### LINKED ISSUE #259: d.ts problem at VS Code
Hey, 

Is it possible to fix this issue

https://github.com/Microsoft/vscode/issues/54165

Avaq:
Fixes #259

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -51,32 +51,32 @@ declare module 'fluture' {
     (a: A): AwaitingTwo<B, C, R>
   }
 
-  export interface ConcurrentFuture<L, R> {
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -342,16 +342,16 @@ declare module 'fluture' {
 
   }
 
-  export var Future: Fluture
+  export var Future: FutureTypeRep
   export default Future
```
Avaq: It's all about this line here. This used to mean "export the `Future` *interface* as `default`", and now it means "export the `Future` *value* as `default`".

PR #269, opened 2018-07-30
Implement chainRec in terms of chain

Avaq:
Since chain is stack safe by itself, there is no need for a duplicate
implementation of trampolining.


davidchambers:
:clap:

PR #276, opened 2018-08-28
Migrate to mjs

### LINKED ISSUE #274: Fluture with `esm`
I am unable to use Fluture with [`esm`](https://github.com/standard-things/esm/).

```js
// @file index.js
import {Future} from 'fluture';
```

```cli
$ node -r esm index.js
SyntaxError:
```

Avaq:
This is (kind-of) the long-awaited followup of https://github.com/fluture-js/Fluture/commit/96e1b95a5698c79cc536343a3945ea9ad863d4c7!

This change allows Node 9+ and [`esm`](https://github.com/standard-things/esm) users to `import {named} from 'fluture'`. However, unlike what I said in 96e1b95a5698c79cc536343a3945ea9ad863d4c7, I will continue to include a `.js` build in the published package until all major tooling has caught up with the new `.mjs` loader. The purpose of this change then becomes solely to support esm *properly* - something I thought was already the case after 96e1b95a5698c79cc536343a3945ea9ad863d4c7.

---

- All `.js` file extensions renamed to `.mjs`
- Some import statements changed to comply to Node 10's [cjs interop](https://nodejs.org/api/esm.html#esm_interop_with_existing_modules) (where before they relied on an [esm option](https://github.com/standard-things/esm#options) for easier interop).
- Sadly, Mocha cannot natively load `.mjs` files, nor can Istanbul. So the unit tests will now be run using `esm` with the approach suggested in https://github.com/standard-things/esm/issues/498#issuecomment-403737113.

Closes #274


Avaq:
I've labelled this a **breaking** change, because the primary file names have changed. For users who would have had something like `import 'fluture/index.mjs.js'`, this will break their builds.

Generally we don't consider changes in file names breaking, unless those names were documented. But I think for the main files a major version bump might be wise.

davidchambers:
> But I think for the main files a major version bump might be wise.

You're unlikely to run out of major version numbers. :smile:

PR #277, opened 2018-08-29
Drop support for "primitive" dispatching

Avaq:
A while ago, I added sanctuary-type-classes as a dependency to Fluture,
primarily to gain its fantastic toString implementation. As a
side-effect, Fluture got to use its dispatcher implementations for ap,
map, bimap, chain, and alt.

Now that sanctuary-show has the toString logic, there is little to gain
from including the entirety of sanctuary-type-classes for dispatching
to FL methods.

The gains were:

- Some of its logic did not have to be reimplemented. Though as shown
  by this diff, not that much.
- Fluture dispatchers could be used on primitive types
  (like Fluture.map(f, [1, 2, 3])). I believe this functionality was
  rarely, if ever, used. Furthermore, TypeScript users were unable to
  use it out of the box because of the limitations of TypeScript.

The losses however:

- A fairly big increase of the bundle size, which as of recent I have
  been working to reduce.
- A minor hit on the performance of these dispatchers.
- The management cost of an additional dependency.
- Breaking changes to sanctuary-type-classes primitive dispatching would
  propagate to a breaking change in Fluture which was confusing to users
  (see https://git.io/fAGsJ#issuecomment-355616168 for example).

### DIFF
--- a/src/internal/const.mjs
+++ b/src/internal/const.mjs
@@ -1,9 +1,10 @@
 export var FL = {
-  map: 'fantasy-land/map',
+  alt: 'fantasy-land/alt',
+  ap: 'fantasy-land/ap',
   bimap: 'fantasy-land/bimap',
   chain: 'fantasy-land/chain',
   chainRec: 'fantasy-land/chainRec',
-  ap: 'fantasy-land/ap',
+  map: 'fantasy-land/map',
   of: 'fantasy-land/of',
   zero: 'fantasy-land/zero'
 };
--- a/src/internal/predicates.mjs
+++ b/src/internal/predicates.mjs
@@ -3,7 +3,7 @@ export function isFunction(f){
 }
 
 export function isThenable(m){
-  return m instanceof Promise || Boolean(m) && isFunction(m.then);
+  return m instanceof Promise || m != null && isFunction(m.then);
 }
 
 export function isBoolean(f){
--- a/src/internal/predicates.mjs
+++ b/src/internal/predicates.mjs
@@ -1,3 +1,5 @@
+import {FL} from './const';
+
 export function isFunction(f){
   return typeof f === 'function';
 }
@@ -28,4 +30,28 @@ export function isIterator(i){
 
 export function isArray(x){
   return Array.isArray(x);
+}
+
+export function hasMethod(method, x){
+  return x != null && isFunction(x[method]);
+}
+
+export function isFunctor(x){
+  return hasMethod(FL.map, x);
+}
+
+export function isAlt(x){
+  return isFunctor(x) && hasMethod(FL.alt, x);
+}
+
+export function isApply(x){
+  return isFunctor(x) && hasMethod(FL.ap, x);
+}
+
+export function isBifunctor(x){
+  return isFunctor(x) && hasMethod(FL.bimap, x);
+}
+
+export function isChain(x){
+  return isApply(x) && hasMethod(FL.chain, x);
 }
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -26,7 +26,9 @@
     "camelcase": 0,
     "indent": 0,
     "func-name-matching": 0,
-    "no-console": 2
+    "no-console": 2,
+    "eqeqeq": [2, "smart"],
+    "no-eq-null": 0
   },
   "overrides": [
     {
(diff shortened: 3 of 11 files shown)


Avaq:
Performance benchmarking results:

```console
$ npm run bench -- --match '*def.transform*'

index.cjs.mjs → index.js...
created index.js in 311ms
Running benchmarks: fluture...

┌─────────────────────┬────────────────────────────┬─────────────────────────────┬────────┬──────────┬───┐                                                                                     
│ suite               │ Old                        │ New                         │ diff   │ change   │ α │
├─────────────────────┼────────────────────────────┼─────────────────────────────┼────────┼──────────┼───┤
│ def.transform.ap    │ 540,492 Hz ±0.34% (n 94)   │ 7,862,581 Hz ±0.27% (n 94)  │ 087.1% │ +1354.7% │ ✓ │
├─────────────────────┼────────────────────────────┼─────────────────────────────┼────────┼──────────┼───┤
│ def.transform.map   │ 2,206,611 Hz ±0.84% (n 93) │ 16,387,149 Hz ±0.68% (n 94) │ 076.3% │ +642.6%  │ ✓ │
├─────────────────────┼────────────────────────────┼─────────────────────────────┼────────┼──────────┼───┤
│ def.transform.chain │ 776,211 Hz ±1.26% (n 91)   │ 8,923,991 Hz ±0.43% (n 95)  │ 084.0% │ +1049.7% │ ✓ │
└─────────────────────┴────────────────────────────┴─────────────────────────────┴────────┴──────────┴───┘
```

src/internal/predicates.mjs
```diff
@@ -29,3 +31,27 @@ export function isIterator(i){
 export function isArray(x){
   return Array.isArray(x);
 }
+
+export function hasMethod(method, x){
+  return Boolean(x) && typeof x[method] === 'function';
```
davidchambers: Why `Boolean(x)` rather than `x != null`? It seems wrong to me:

```javascript
> (function (method, x) { return Boolean (x) && typeof x[method] === 'function'; } ('concat', 'xyz'))
true

> (function (method, x) { return Boolean (x) && typeof x[method] === 'function'; } ('concat', ''))
false
```

src/internal/predicates.mjs
```diff
@@ -29,3 +31,27 @@ export function isIterator(i){
...
+}
+
+export function isFunctor(x){
+  return hasMethod(FL.map, x);
```
davidchambers: What about `[]`, `{}`, `Math.abs`, and so on?

src/internal/const.mjs
```diff
@@ -1,9 +1,10 @@
...
+  alt: 'fantasy-land/alt',
+  ap: 'fantasy-land/ap',
   bimap: 'fantasy-land/bimap',
   chain: 'fantasy-land/chain',
   chainRec: 'fantasy-land/chainRec',
-  ap: 'fantasy-land/ap',
+  map: 'fantasy-land/map',
   of: 'fantasy-land/of',
   zero: 'fantasy-land/zero'
 };
```
davidchambers: Consider making this mapping exhaustive even if some of the values are never referenced.

src/internal/predicates.mjs
Avaq: That is a good point. I've been using `Boolean` to check if something is "accessible", but I hadn't considered strings, numbers and Booleans.

davidchambers: `null` and `undefined` are the only “inaccessible” values, as far as I'm aware.

src/internal/predicates.mjs
Avaq: Sadly, those don't have Fantasy Land implementations, and are exactly what I'm dropping support for with this commit (referring to it as "primitive dispatching"). Users who want to treat these values as Functors can use (and probably already are using) Ramda or Sanctuary. `map` and friends in Fluture can be thought of as operating on Futures and ConcurrentFutures (although they can work with any FL type).

Fluture currently supports it through sanctuary-type-classes, but at too great a cost. I am trying to shrink the Fluture core down to the essentials, and would like to [integrate with Sanctuary at a higher level](https://github.com/fluture-js/Fluture/projects/3#card-11455088).

src/internal/predicates.mjs
```diff
@@ -1,9 +1,15 @@
+import {FL} from './const';
+
+export function isAccessible(x){
+  return x !== null && x !== undefined;
```
davidchambers: You could use `x != null` here. This is such a common idiom that I don't consider `isAccessible` warranted.

PR #279, opened 2018-08-30
Replace Future.finally

Avaq:
### Remove Future.finally and its aliases

The finally function has long been surpassed by Future.hook, and has
several problems:

1. An undocumented behaviour where it will run *and cancel* your
   cleanup Future upon cancellation. This once was intended, and
   copied to Future.hook, but then it was decided to be bad behaviour
   and removed from Future.hook. It stuck around in Future.finally.
2. It doesn't clean up when the Future crashes.
3. It's not very useful for any other kind of control flow.

### Add Future.assume to fill the gap that Future.finally left

I've personally been using Future.finally *only* to shorten this:

    someFuture
      .chain(_ => Future.reject('Kaputt'))
      .chainRej(_ => Future.reject('Kaputt'))

To this:

    someFuture.finally(Future.reject('Kaputt'))

Now that finally is gone, I'll replace it by a function which does this
job, and does it better (it works for resolved Futures as well).


Avaq:
I'm going to fix the `finally` semantics instead of removing it entirely.

PR #282, opened 2018-08-31
Add the Future#pipe method

Avaq:
For those who prefer the fluent method API, this function grants
third-party functions similar privileges to the methods included in
Flutre by default, by allowing those functions to be called on the
Future without breaking the fluent method chain.

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,6 +169,7 @@ for sponsoring the project.
 
 <details><summary>Transforming and combining Futures</summary>
 
+- [`pipe`: Apply a function to a Future in a fluent method chain](#pipe)
 - [`map`: Synchronously process the success value in a Future](#map)
 - [`bimap`: Synchronously process the success or failure value in a Future](#bimap)
 - [`chain`: Asynchronously process the success value in a Future](#chain)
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -1604,6 +1606,30 @@ will be ignored.
...
+<details><summary><code>pipe :: Future a b ~> (Future a b -> c) -> c</code></summary>
+
+```hs
+pipe :: Future a b ~> (Future a b -> c) -> c
+```
+
+</details>
+
+A method available on all Futures to allow arbitrary functions to be included
+in a fluent-style method chain.
```
davidchambers: I understand what you mean by _arbitrary functions_, but the wording is imprecise. You're referring to functions not defined by Fluture as arbitrary, but readers may interpret _arbitrary_ as synonymous with _any_. As the function must accept a future as input, _arbitrary_ is too broad, I think.

Avaq: I cannot quite think of a better word, what do you think of *arbitrary functions over Futures*?

davidchambers: That's perfect!

Avaq: :zap:

PR #283, opened 2018-09-01
Async stack traces and debug mode

Avaq:
Changes

- New debugging utilities allow Fluture to take snapshots of the stack.
- All Future interpretors now call the recover function with additional
  information on top of the crash value when they encounter an exception.
  This information includes stack snapshots taken at potentially
  different tick.
- Abstractions on top of these low-level interpretors use the additional
  information to form extended stack traces for the errors they create.
- A new public 'debugMode' function allows users to control whether the
  above features are activated.

----

You can test this yourself by installing this branch:

```console
$ npm i fluture-js/fluture#avaq/debug
```

Then write an `.mjs` file using ES6 imports from `fluture` and run it with:

```console
$ node --experimental-modules --no-warnings the-file.mjs
```

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -28,7 +28,8 @@
     "func-name-matching": 0,
     "no-console": 2,
     "eqeqeq": [2, "smart"],
-    "no-eq-null": 0
+    "no-eq-null": 0,
+    "no-param-reassign": 0
   },
   "overrides": [
     {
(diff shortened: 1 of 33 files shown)


Avaq:
Currently the `debugMode` api is global. I've been thinking about an elegant way around this, and I think I have a solution. Instead of controlling debug mode globally, one would control it per Future instance. When an instance is set to debug mode, all futures interpreted by its interpreter would also run in debug mode, so you'll only need to call it once on your top-level Future instance.

It would look like this:

```hs
debug :: Future a b -> Future a b
```

```js
import {debug} from 'fluture';
import {doAsyncStuff} from './src';

debug(doAsyncStuff()).fork(console.error, console.log)
```

Avaq:
> I think I have a solution

Never mind. This idea would only work if debugMode didn't also affect the various Future constructors. :\

PR #299, opened 2018-11-28
Fix url.

fr-sutiwo:
The order of `Object.create` and `Object.assign` is opposite.


Avaq:
Oh! Thanks. Hahaha. That was silly.

Issue #89, opened 2017-04-19
Implement Alternative using race/never


Avaq:
Now that `race` has a special case for `Never` instances, we can implement lawful Alternative using `race` and `never`. This would mean that conversion to the `ConcurrentFuture` would *only* be necessary to get parallel `ap` behavior.

safareli:
https://github.com/fluture-js/concurrify/issues/3#issuecomment-295275924

Avaq:
These ideas originated in a conversation with you on Gitter:

> I've been lead to another question by reading the (beautiful) source-code of [Creed](https://github.com/briancavalier/creed). Creed implements chain-derived (sequential) `ap` *and* parallel `alt` on the same type. Why did we consider this impossible again?
> -- @avaq

> there is no "clear" answer to that, technically you could do it and it will be lawful
briancavalier/creed#92
fantasyland/fantasy-land#203
> but Monadic structure could be MonadPlus (or MonadOr or MonadZero) and i didn't went there farther because there is still debate on it so if it's finalised and we decide to "port" it to FL then such alternative instance could be invalid if structure is also made MonadPlus...
> -- @safareli 

> Ah. So even in Haskell they're not so sure. o.o
> Because in that case I could've implemented concurrify as merely Apply, and extend it as such:
> ```js
> const ConcurrentFuture = concurrify(Future, myParalellAp);
> ConcurrentFuture[FL.of] = function(x){ return ConcurrentFuture(Future[FL.of](x)) }
> ConcurrentFuture[FL.zero] = function(){ return ConcurrentFuture(Future[FL.zero]()) }
> ConcurrentFuture.prototype[FL.alt] = function(b){ return this[FL.alt](b) }
> ```
> -- @avaq

----

It seems that at the time, you were unsure if Monad with Alternative is lawful. It seems that @briancavalier has chosen for the more convenient approach of having both instances on his structure in briancavalier/creed#92. If no new insights have come to light in the mean time, I would be inclined to proceed with this idea. Since it seems to be a polarizing subjects, though, I will not push this for the `6.0` release. I'm also curious to learn whether Brian Cavalier's ideas about this are still the same (hence the ping).

briancavalier:
> I've been lead to another question by reading the (beautiful) source-code of Creed

Thank you, @Avaq!

> the more convenient approach of having both instances

[...]

PR #301, opened 2018-12-01
Implement Alt instance on Future

Avaq:
As [discussed on Gitter](https://gitter.im/fluture-js/Fluture?at=5bfdb54c958fc53895fc9904) with @masaeedu and @syaiful6, implementing sequential Alt on Future is the better option compared to implementing parallel Alternative, which was also discussed in #89.


README.md
```diff
@@ -182,7 +183,7 @@ for sponsoring the project.
 - [`fold`: Coerce success and failure values into the same success value](#fold)
 - [`ap`: Combine the success values of multiple Futures using a function](#ap)
 - [`and`: Logical *and* for Futures](#and)
-- [`or`: Logical *or* for Futures](#or)
+- [`alt`: Logical *or* for Futures](#alt)
```
masaeedu: Is deleting this a breaking change? If so, why not leave both?

Avaq: I did leave both, I left "or" as an alias to alt (I added mention about this in the documentation of `alt`), it's just that I want "alt" to be leading.

PR #336, opened 2019-05-07
Fix the typescript signature for fold in stable

### LINKED ISSUE #334: Fold typescript definition uses potentially pointless generic L
Correct me if I'm wrong, but fold is defined as:

```hs
fold  :: Future a b ~> (a -> c, b -> c) -> Future d c
```

However in typescript (index.d.ts) fold is defined as:

```ts
fold<L, RB>(lmapper: (reason: L) => RB, rmapper: (value: R) => RB): FutureInstance<L, RB>
```

or

```hs
fold  :: Future a b ~> (a -> c, b -> c) -> Future a c
```

right? I believe the correct definition should be:

```ts
fold<RL, RB>(lmapper: (reason: L) => RB, rmapper: (value: R) => RB): FutureInstance<RL, RB>
//or
fold<RB>(lmapper: (reason: L) => RB, rmapper: (value: R) => RB): FutureInstance<L, RB>
//or
fold<RB>(lmapper: (reason: L) => RB, rmapper: (value: R) => RB): FutureInstance<void, RB>
```

Personally I would prefer the latter one

Avaq:
Closes #334

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -97,7 +97,7 @@ declare module 'fluture' {
     finally(cleanup: FutureInstance<L, any>): FutureInstance<L, R>
 
     /** Fold both branches into the resolution branch. See https://github.com/fluture-js/Fluture#fold */
-    fold<L, RB>(lmapper: (reason: L) => RB, rmapper: (value: R) => RB): FutureInstance<L, RB>
+    fold<LA, RA, LB, RB>(lmapper: (reason: LA) => RB, rmapper: (value: RA) => RB): FutureInstance<LB, RB>
 
     /** Fork the Future into the two given continuations. See https://github.com/fluture-js/Fluture#fork */
     fork(reject: RejectFunction<L>, resolve: ResolveFunction<R>): Cancel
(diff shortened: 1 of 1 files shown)


Avaq:
@bagdonas If you're okay with using the beta version of Fluture, we can close this PR. Otherwise we can release it as a patch into v11.

bagdonas:
@Avaq A patch into v11 would break existing code, I'm ok with current behavior until v12, however I disagree with the new definition.

In `fold<LA, RA, LB, RB>(lmapper: (reason: LA) => RB, rmapper: (value: RA) => RB): FutureInstance<LB, RB>` `LA` and `RA` are unnecessary since they are already defined in `FutureInstance<L, R>` as `L` and `R`

I believe better definition would be:
```
fold<LB, RB>(lmapper: (reason: L) => RB, rmapper: (value: R) => RB): FutureInstance<LB, RB>
```

If there's a need for all 4 generics, there's always an option to overload the method:
```
fold<LB, RB>(lmapper: (reason: L) => RB, rmapper: (value: R) => RB): FutureInstance<LB, RB>
fold<LA, RA, LB, RB>(lmapper: (reason: LA) => RB, rmapper: (value: RA) => RB): FutureInstance<LB, RB>
```

And if we're overloading the method, there might as well be an option to pass only the RB generic 😃 
```
fold<RB>(lmapper: (reason: L) => RB, rmapper: (value: R) => RB): FutureInstance<never, RB>
fold<LB, RB>(lmapper: (reason: L) => RB, rmapper: (value: R) => RB): FutureInstance<LB, RB>
fold<LA, RA, LB, RB>(lmapper: (reason: LA) => RB, rmapper: (value: RA) => RB): FutureInstance<LB, RB>
```

Avaq:
> `LA` and `RA` are unnecessary since they are already defined in `FutureInstance<L, R>`

Oh, indeed. My mistake. Am I correct to assume, however, that your disagreement is only with the *method* definition of `fold`?

Note that the method has been removed in v12. Only the function remains. So we wouldn't need to make any changes to the master branch, correct?

Avaq:
> ```ts
> fold<RB>(lmapper: (reason: L) => RB, rmapper: (value: R) => RB): FutureInstance<never, RB>
> ```

This is in essence [how `fold` was typed](https://github.com/fluture-js/Fluture/pull/230/files#diff-b52768974e6bc0faccb7d4b75b162c99L221) before #230.

The problem with `never` is that it's not compatible with other types (it not a type variable), so after folding, a `chain` would get TypeScript to complain. I think `unknown` would be a better fit. See #337.

Avaq:
> A patch into v11 would break existing code, I'm ok with current behavior until v12

Good point. I'm closing this. Don't feel discouraged to share your thoughts about #337, though! :)

bagdonas:
Yes, I'm only concerned with fold. Old definition was fine except that `L` from `fold<L, RB>` overshadowed `L` from `FutureInstance<L, R>` and was unnecessary.

I wasn't aware that v12 is purely functional 😞

Avaq:
I'm sorry to find out the purely functional API saddens you. Note that the `pipe` method remains, so there are [numerous alternatives to using the method API](https://gist.github.com/Avaq/ee2c6c819db4c37258e9a226e6380a38#fluent-method-api-removed).

The advantages of decreased API surface, API consistency, increased source modularity, and better static analysis (eg. for tree shaking), I feel outweigh the convenience of a fluent method API.

PR #342, opened 2019-05-19
Refactor code to have all core building blocks in one file

### LINKED ISSUE #318: Tree shaking for Fluture is not as potent as it could be in theory
For some reason, most of the Fluture code-base is included any [rollup](https://rollupjs.org)-created build, independent of which functions are loaded (and even from which files they are loaded).

I've pushed [some experiments in the `avaq/treeshaking` branch](https://github.com/fluture-js/Fluture/tree/avaq/treeshaking/rollup) that I was using to identify the cause, but I've halted work in order to focus my attention elsewhere.

Avaq:
This indirectly closes #318, as demonstrated by the [treeshaking branch](https://github.com/fluture-js/Fluture/tree/avaq/treeshaking/rollup). Moving all code that touches the Future prototype into one file means that no other files are required for their "side effects", allowing users to set the `treeshake.moduleSideEffects` to `false` when tree-shaking Fluture.

Besides solving the tree-shaking issue, although this moves more code into a single file, I find that it's clears up the structure of the code base. Instead of implicit circular dependencies, we have no circular dependencies. The entire Fantasy Land Future type is defined in a single file, and all other files contain the functions surrounding this type.


dicearr:
```diff
- treeshare.moduleSideEffects
+ treeshake.moduleSideEffects
```

PR #345, opened 2019-06-10
Add support for experimental modules in Node 12

Avaq:
Node 12 wants modules to explicitly state the file extension in import
paths, so this commit adds file extensions to every import statement.


Avaq:
:warning: This might harm users who consume the Fluture codebase using a transpiler: https://github.com/nodejs/modules/issues/323#issuecomment-490291757

Avaq:
>> From a transpiler’s or to-JS compiler’s perspective, what new features are needed?
>
> None. Transpilers can add the ability to add extensions at compile time, so a specifier like './file' can be rewritten to './file.js' during compilation along with whatever else is getting converted by the transpiler. This would provide similar UX without needing the flag, if the user didn’t want to either just type the extension or use the flag.
> <sub>-- https://github.com/nodejs/modules/issues/323#issuecomment-492813819</sub>

That makes it sound like transpilers will have no issues with the change.

Avaq:
Actually, there's more work to be done here. It seems that Node 12 has dropped support for modules that export *both* `.js` and `.mjs` "main" files.

> Currently, it is not possible to create a package that can be used via both require(‘pkg’) and import ‘pkg’. -- https://medium.com/@nodejs/announcing-a-new-experimental-modules-1be8d2d6c2ff#576c

This means that, although this patch fixes `import {...} from 'fluture/index.mjs'`, `import {...} from 'fluture'` will **always** throw an Error, because Node 12 will resolve that to the CJS module (or throw an Error? https://github.com/nodejs/node/issues/28160), which doesn't have named exports.

Avaq:
The same section also asks the following of its readers:

> Please do not publish any ES module packages intended for use by Node.js until this is resolved.

So it's probably best to wait until this capability is reintroduced to Node. In the meantime, I think no harm is done by merging this PR.

Avaq:
This PR breaks the `npm run test:coverage` script due to what seems like an incompatibility between Node 12, esm, and/or nyc. I think it will have to wait.

Avaq:
> This PR breaks the `npm run test:coverage`

Turns out it's not this PR, but Node 12 in general.

PR #356, opened 2019-07-07
Improve the experience of converting a Future to JSON

### LINKED ISSUE #350: Fluture's exceptions break various error handlers
For the simple reason that it contains recursive data structures; namely Fluture's linked list.

https://github.com/fluture-js/Fluture/blob/2e455577c5b7250090b07e2bc35c91a4f316ddad/src/internal/list.mjs#L1-L6

Error handlers known to choke on this:

- JSVerify's error display logic
- AWS Lambda error display logic

Avaq:
Since casting a Future to JSON doesn't really make sense, I've decided
to implement it as returning a kind of AST. This AST can be useful for
debugging. Since most of the scenarios where a Future is being turned
into JSON is during error handling, returning a value useful for
debugging seemed the most appropriate.

Closes #350

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,6 +136,7 @@ for sponsoring the project.
 - [How cancellation works](#cancellation)
 - [On stack safety](#stack-safety)
 - [Debugging with Fluture](#debugging)
+- [Casting Futures to String](#casting-futures-to-string)
 - [Usage with Sanctuary](#sanctuary)
 - [Using multiple versions of Fluture](#casting-futures)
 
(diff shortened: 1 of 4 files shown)


Avaq:
An alternative to the AST is to have JSON like:

```js
{
  "fluture/Future@5": "ap (resolve (22)) (map (a => b => a + b) (resolve (20)))"
}
```

Yet another alternative is to have the `toJSON` be only present on `list`, without the custom `toJSON` on interpreters and transformations. Then the output of `stringify(eventualAnswer)` would be this:

```js
{
  "context": [],
  "$1": {
    "context": [],
    "$1": 20
  },
  "$2": [
    {
      "context": [],
      "$1": {
        "context": [],
        "$1": 22
      }
    },
    {
      "context": []
    }
  ]
}
```

PR #363, opened 2019-08-21
Add guards to parallel to prevent continuation after cancel

### LINKED ISSUE #362: Error thrown when using same cached Future within Future.parallel
Hi, I'm running into a possible bug here when referencing a cached future more than once from within a call to `Future.parallel` and then adding some sort of continuation onto it (e.g. `map`, `mapRej`, etc.).

```js
const noop = () => {};

const cachedFuture = Future.cache(Future.rejectAfter(1, new Error));

Future
	.parallel(Infinity, [cachedFuture, cachedFuture])
	.map(() => true)
	.fork(noop, noop);
```

This code will throw `TypeError: Cannot read property 'rejected' of null`.

Expected behavior is that this does not throw an error. What's interesting to me is that if you remove the `.map` call or use a cached future that is rejected right away i.e. `Future.cache(Future.reject(new Error))` then it'll work.

Avaq:
Fixes #362

/cc @jonwhelan

### DIFF
--- a/src/parallel.mjs
+++ b/src/parallel.mjs
@@ -19,10 +19,13 @@ Parallel.prototype._interpret = function Parallel$interpret(rec, rej, res){
 
   var _futures = this._futures, _length = this._length, _max = this._max;
   var cancels = new Array(_length), out = new Array(_length);
-  var cursor = 0, running = 0, blocked = false;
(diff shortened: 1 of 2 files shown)


Avaq:
Not sure what Codecov is on about there. Coverage remained at 100%.

Issue #337, opened 2019-05-07
Stop using user-assignable type generics in the types of functions that don't do anything with them


Avaq:
This is an alternative to #230 made possible by the [introduction of the `unknown` type](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-0.html#new-unknown-top-type).

See also https://github.com/fluture-js/Fluture/issues/334#issuecomment-490139148.
See also https://github.com/fluture-js/Fluture/issues/370#issuecomment-533226653.

Avaq:
TypeScript 3.4 introduces [higher order type inference from generic functions](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#higher-order-type-inference-from-generic-functions). This might make the need for using anything but an undetermined type variable/generic unnecessary. I should do some experiments with this.

Avaq:
<sup>In response to https://github.com/fluture-js/Fluture/issues/370#issuecomment-533715945</sup>

I tried a simple thing myself, and it seems `never` and `unknown` have the same issues, but in opposite orders.

### With `never`/`undefined`

![latest-screenshot](https://user-images.githubusercontent.com/1217745/65394927-d1205980-dd93-11e9-86a7-5678ddabfbc1.png)
![latest-screenshot](https://user-images.githubusercontent.com/1217745/65394952-3aa06800-dd94-11e9-9e9c-d8f5a2647522.png)

Here the problem is that we're not allowed to __assign anything to `never`__ (the same goes for `undefined`). So when we're trying to assign `Future<string, never>` to `Future<never, number>`, it breaks on the *rejection branch*.

### With `unknown`

![latest-screenshot](https://user-images.githubusercontent.com/1217745/65394921-b9e16c00-dd93-11e9-95a6-83ac3118fead.png)

Here the problem is the other way around: we're not allowed to __assign `unknown` to anything__. So when we're trying to assign `Future<string, unknown>` to `Future<unknown, number>`, it breaks on the *resolution branch*.

----

In conclusion, it seems that neither `never` nor `unknown` possess the desired behaviour.

Avaq:
It seems that `never` has the behaviour we want, but only when `chain` uses a union to define the type of the left branch. The downside is that this will allow users to mix their types through usage of `chain`.

edgarjrg:
Yes i agree that neither of the types work right now.

PR #373, opened 2019-09-18
Move the declaration of generics deeper into type signatures

### LINKED ISSUE #372: Move TypeScript generics as deeply into function signatures as possible
> What about making `fold`
> ```ts
> export function coalesce <LA, RB>(lmapper: (left: LA) => RB): <RA>(rmapper: (right: RA) => RB) => <LB>(source: FutureInstance<LA, RA>) => FutureInstance<never, RB>
> ```
> I suppose that should be because some throw, like in the case of
> > See fluture-js/Fluture#331. In order to call promise on a Future, the rejection branch must now be compatible with type Error, and thrown exceptions are now merged with the rejection branch, instead of being rethrown.
> 
> but not sure if that's a problem of `fold`
> 
> Also in general add the generic `<T>` as late as possible has proven to be more effective, with the `fluture@12.0.0-beta.3` i needed to type way more functions, specially fold and bimap. when i change `bimap` to 
> ```ts
> export function bimap<LA, LB>(lmapper: (reason: LA) => LB): <RA, RB>(rmapper: (value: RA) => RB) => (source: FutureInstance<LA, RA>) => FutureInstance<LB, RB>
> ```
> ts infers way better the bimap type.

_Originally posted by @edgarjrg in https://github.com/fluture-js/Fluture/issues/370#issuecomment-532073361_

edgarjrg:
To close #372

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -92,7 +92,7 @@ declare module 'fluture' {
   export function attemptP<L, R>(fn: () => Promise<R>): FutureInstance<L, R>
 
   /** Map over both branched of the given Bifunctor at once. See https://github.com/fluture-js/Fluture#bimap */
-  export function bimap<LA, LB, RA, RB>(lmapper: (reason: LA) => LB): (rmapper: (value: RA) => RB) => (source: FutureInstance<LA, RA>) => FutureInstance<LB, RB>
+  export function bimap<LA, LB>(lmapper: (reason: LA) => LB): <RA, RB>(rmapper: (value: RA) => RB) => (source: FutureInstance<LA, RA>) => FutureInstance<LB, RB>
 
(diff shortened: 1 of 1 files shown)


edgarjrg:
@Avaq I started with `bimap` and `fold` and , i'll check others, if you have any other let me know.

Avaq:
Some others where I think this may be beneficial are `bimap`, `fork` and `forkCatch`. But I think once #337 is resolved, many other functions can benefit:

- `after`
- `rejectAfter`
- `ap`
- `both`
- `hook`
- `lastly`
- `map`
- `mapRej`
- `parallel`

Avaq:
Let's do the remainder of the function which we can without fixing #337, so we can get this merged. It seems #337 needs a bit more work.

Issue #371, opened 2019-09-16
Querion how to chain with different left type


edgarjrg:
Hi ! 

This tests is successful for me
```
test('chain with different left type', () => {
    fork(x => expect(x).toEqual(3))(fail)(chain<boolean, any, any>(x => reject(false))(reject(3)))
  })
```
However i have to ignore the types otherwise i would get
```
Argument of type 'FutureInstance<number, any>' is not assignable to parameter of type 'FutureInstance<boolean, any>'.
  Type 'number' is not assignable to type 'boolean'.ts(2345)
```
What is the right method that has this type signature
```ts
export function chain?<LA, RA, LB, RB>(mapper: (value: RA) => FutureInstance<LB, RB>): (source: FutureInstance<LA, RA>) => FutureInstance<LB, RB>

export function chainRej?<LA, RA, LB, RB>(mapper: (value: LA) => FutureInstance<LB, RB>): (source: FutureInstance<LA, RA>) => FutureInstance<LB, RB>
```

That's pretty much how i chain `Promises`, but maybe Futures are supposed to be though or combined differently.

edgarjrg:
As it may have not applied the `mapper` at all if rejected, actually i think it may be more like this:

```ts
export function chain?<LA, RA, LB, RB>(mapper: (value: RA) => FutureInstance<LB, RB>): (source: FutureInstance<LA, RA>) => FutureInstance<LA, never> | FutureInstance<LB, RB>
```
That resolves to 
```ts
export function chain?<LA, RA, LB, RB>(mapper: (value: RA) => FutureInstance<LB, RB>): (source: FutureInstance<LA, RA>) => FutureInstance<LA | LB, RB>
```

Promises never deal with how the error is typed as is always `any`, but actually i think this is the more accurate representation, but again, my lack of knowledge of the Future type might be in the way. I've yet not seen any library that deals with different types in the left branch, is always one generic type `E` that never changes.

Avaq:
> how to chain with different left type

It is not allowed to `chain` two different types of Future: `(Future Number) a` is not the same Monad as `(Future Boolean) a`.

The problem is that without Type Checking, we can never ensure this constraint using just plain JavaScript. So only TypeScript users are aware of the constraint.

The solution is to make the types match before you `chain`:

```js
const boolToInt = b => b ? 0 : 1;

fork (x => expect(x).toEqual(3))
     (fail)
     (chain(x => mapRej(boolToInt)(reject(false)))
[...]

PR #374, opened 2019-09-25
Improve TypeScript type inference

### LINKED ISSUE #337: Stop using user-assignable type generics in the types of functions that don't do anything with them

### LINKED ISSUE #370: types of fold
Hi !

The type of fold (or coalesce as proposed) is

```ts
export function fold<LA, RA, LB, RB>(lmapper: (left: LA) => RA, rmapper: (right: RA) => RB, source: FutureInstance<LA, RA>): FutureInstance<LB, RB>
```

I think it should be 
```ts
export function fold<LA, RA, RB>(lmapper: (left: LA) => RB, rmapper: (right: RA) => RB, source: FutureInstance<LA, RA>): FutureInstance<never, RB>
```

ok, maybe the never shouldn't be there, but the `lmapper` seems wrong.

### LINKED ISSUE #372: Move TypeScript generics as deeply into function signatures as possible

Avaq:
After the discussions had in #337 and #370, I think our problems cannot be solved without compromising the strictness of the `chain` and `chainRej` functions. This PR makes the following changes to get type inference to work well:

- Revert #230 and reintroduce the usage of the 'never' type
- Relax the signature of chain and chainRej to allow TypeScript to
  derive types when merging a Future of a never with a Future of a
  real type. Closes #337 and closes #370.
- Move generic types closer to where they are used in the function
  signature. Closes #372 and closes #373 through supersession.

/cc @edgarjrg, @tetsuo

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -63,12 +63,12 @@ declare module 'fluture' {
     'fantasy-land/map'<RB>(mapper: (value: R) => RB): FutureInstance<L, RB>
     'fantasy-land/alt'(right: FutureInstance<L, R>): FutureInstance<L, R>
     'fantasy-land/bimap'<LB, RB>(lmapper: (reason: L) => LB, rmapper: (value: R) => RB): FutureInstance<LB, RB>
-    'fantasy-land/chain'<RB>(mapper: (value: R) => FutureInstance<L, RB>): FutureInstance<L, RB>
+    'fantasy-land/chain'<LB, RB>(mapper: (value: R) => FutureInstance<LB, RB>): FutureInstance<L | LB, RB>
 
   }
 
   /** Creates a Future which resolves after the given duration with the given value. See https://github.com/fluture-js/Fluture#after */
-  export function after<L, R>(duration: number): (value: R) => FutureInstance<L, R>
+  export function after(duration: number): <R>(value: R) => FutureInstance<never, R>
 
   /** Logical and for Futures. See https://github.com/fluture-js/Fluture#and */
   export function and<L, R>(left: FutureInstance<L, R>): (right: FutureInstance<L, any>) => FutureInstance<L, R>
@@ -80,10 +80,10 @@ declare module 'fluture' {
   export function alt<L, R>(left: ConcurrentFutureInstance<L, R>): (right: ConcurrentFutureInstance<L, R>) => ConcurrentFutureInstance<L, R>
 
   /** Apply the function in the left Future to the value in the right Future. See https://github.com/fluture-js/Fluture#ap */
-  export function ap<L, RA, RB>(value: FutureInstance<L, RA>): (apply: FutureInstance<L, (value: RA) => RB>) => FutureInstance<L, RB>
+  export function ap<L, RA>(value: FutureInstance<L, RA>): <RB>(apply: FutureInstance<L, (value: RA) => RB>) => FutureInstance<L, RB>
 
   /** Apply the function in the left ConcurrentFuture to the value in the right ConcurrentFuture. See https://github.com/fluture-js/Fluture#ap */
-  export function ap<L, RA, RB>(value: ConcurrentFutureInstance<L, RA>): (apply: ConcurrentFutureInstance<L, (value: RA) => RB>) => ConcurrentFutureInstance<L, RB>
+  export function ap<L, RA>(value: ConcurrentFutureInstance<L, RA>): <RB>(apply: ConcurrentFutureInstance<L, (value: RA) => RB>) => ConcurrentFutureInstance<L, RB>
 
   /** Create a Future which resolves with the return value of the given function, or rejects with the error it throws. See https://github.com/fluture-js/Fluture#attempt */
   export function attempt<L, R>(fn: () => R): FutureInstance<L, R>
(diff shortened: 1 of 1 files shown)


Avaq:
@davidchambers I asked you for a review because I know you've been working on typings for Sanctuary, and you would've likely encountered the same issues with `chain` for `Either`.

Avaq:
Note that the changes I made to `chain` and `chainRej` is this PR go against things I have said in the past:

- https://github.com/fluture-js/Fluture/issues/371#issuecomment-531704900
- https://github.com/fluture-js/Fluture/pull/221#issuecomment-365541706

tetsuo:
`node(done => open('somefile', 'r', done))` doesn't infer the callback value type since `open` returns nothing we can use as a reference. `typeof open` however can give some information.

so given this signature:

```ts
open(a: string, b: string, cb: (er: Error | null, fd: number) => void): void
```
we can find out its return type using the new `infer` keyword:

```ts
type TypeOf<T extends (a: any, b: any, cb: (er: Error | null, value?: any) => void) => void> = T extends (
  a: any,
  b: any,
  cb: (er: Error | null, value?: infer R) => void
) => void
  ? R
  : any
```

example:

```ts
declare function open(path: string, flags: string | number, callback: (err: Error | null, fd: number) => void): void
type T1 = TypeOf<typeof open>
// number (fd)
```

with that we can infer the callback type:

```ts
import { FutureInstance } from 'fluture'

declare function node<F extends (cb: (er: Error | null, value?: any) => void) => void>(
  f: F
): F extends (cb: (er: infer L, value?: infer R) => void) => void ? () => FutureInstance<NonNullable<L>, R> : () => any
declare function node<F extends (a: any, cb: (er: Error | null, value?: any) => void) => void>(
  f: F
): F extends (a: infer A, cb: (er: infer L, value?: infer R) => void) => void
  ? (a: A) => FutureInstance<NonNullable<L>, R>
  : (a: any) => any
declare function node<F extends (a: any, b: any, cb: (er: Error | null, value?: any) => void) => void>(
  f: F
): F extends (a: infer A, b: infer B, cb: (er: infer L, value?: infer R) => void) => void
  ? (a: A, b: B) => FutureInstance<NonNullable<L>, R>
  : (a: any, b: any) => any
declare function node<F extends (a: any, b: any, c: any, cb: (er: Error | null, value?: any) => void) => void>(
  f: F
): F extends (a: infer A, b: infer B, c: infer C, cb: (er: infer L, value?: infer R) => void) => void
  ? (a: A, b: B, c: C) => FutureInstance<NonNullable<L>, R>
  : (a: any, b: any, c: any) => any

import { open, close } from 'fs'

const t1 = node(open)
// (a: PathLike, b: string | number) => FutureInstance<NodeJS.ErrnoException, number>
const t2 = node(open)(__dirname + '/bla', 'r')
// FutureInstance<NodeJS.ErrnoException, number>
const t3 = node(close)
// (a: number) => FutureInstance<NodeJS.ErrnoException, unknown>
```

rest of the family

```ts

declare function map<A, B>(f: (a: A) => B): <L>(fa: FutureInstance<L, A>) => FutureInstance<L, B>

declare function double(n: number): number

const t4 = map(double)
// <L>(fa: FutureInstance<L, number>) => FutureInstance<L, number>

const t5 = t4(node(open)(__dirname + '/../../package.json', 'r'))
// FutureInstance<NodeJS.ErrnoException, number>

declare function of<A>(a: A): FutureInstance<never, A>

// declare function ap<A, B>(f: (a: A) => B): <L>(fa: FutureInstance<L, A>) => FutureInstance<L, B>
declare function ap<L, A>(fa: FutureInstance<L, A>): <B>(fab: FutureInstance<L, (a: A) => B>) => FutureInstance<L, B>

const t6 = ap(of(42))
// <B>(fab: FutureInstance<unknown, (a: number) => B>) => FutureInstance<unknown, B>

const t7 = t6(of(a => a + 'world'))
// FutureInstance<unknown, string>

const t8 = ap(map(double)(node(open)(__dirname + '/../../package.json', 'r')))(of(a => a + 'world'))
// FutureInstance<NodeJS.ErrnoException, string>
```

Avaq:
Together with some regular TypeScript users I have tried to find a solution which does not compromise on the strictness of `chain`, but have found only solutions whose compromises are worse. I am going to merge this PR.

PR #377, opened 2019-09-27
Remove the option to return void from Future constructor

### LINKED ISSUE #346: Split the Future constructor into one that does cancellation, and one that doesn't
The `Future` function is currently overloaded with the following two signatures:

```hs
Future :: ((a -> Undefined, b -> Undefined) -> Undefined) -> Future a b
Future :: ((a -> Undefined, b -> Undefined) -> () -> Any) -> Future a b
```

The former doesn't take a user cancellation, and the latter does.

Whether or not the user has returned a cancellation function is checked at runtime.

If the user returned anything besides `() -> Any` or `Undefined`, Fluture throws the [ever](https://github.com/fluture-js/Fluture/issues/330) [confusing](https://stackoverflow.com/questions/56351778/using-fluture-with-aws-service) `The computation was expected to return a nullary function or void`.

As this is a common source of confusion, and a deviation from the API's design philosophy of having no overloaded functions, I think it may be worth splitting the function into two.

Avaq:
Closes #346

### DIFF
--- a/README.md
+++ b/README.md
@@ -543,14 +543,20 @@ takes two callbacks. Both are continuations for the computation. The first is
 When the computation is finished (possibly asynchronously) it may call the
 appropriate continuation with a failure or success value.
 
-Additionally, the computation may return a nullary function containing
+Additionally, the computation must return a nullary function containing
 cancellation logic. See [Cancellation](#cancellation).
+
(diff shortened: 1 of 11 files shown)


Avaq:
I rebased this on #368 because it required changes to many test files.

PR #383, opened 2019-10-03
Switch to a CJS-first approach to package distribution

Avaq:
Closes #382 and closes #385 as this is an alternative to both.

This change sets the `main` field in package.json to the *commonjs* version of Fluture. 

#### Consequences

- CommonJS users can load Fluture using simply `require ('fluture')`.
- Module users, including bundlers, can load Fluture using `import {...} from 'fluture/index.mjs'`. Bundlers will do this automatically, guided by the `module` property in package.json.
- Library authors that depend on Fluture will be forced to make the following choice:
    1. Keep their library as CommonJS, not changing anything.
    1. To use `import fluture from 'fluture'`. This will load the CommonJS version of Fluture into their code. This works, but they'll lose the static analysis benefits that come with the modular version.
    1. To use `import {...} from 'fluture/index.mjs'`. This will load the modular version of Fluture into their code. They keep the static analysis benefits, but they lose the ability to transpile this code to CommonJS without having to specify dependency rewrite rules.


README.md
```diff
@@ -53,18 +53,37 @@ On older environments you may need to polyfill one or more of the following
...
+.pipe (Future.fork (console.error) (console.log))
+```
+
 ### EcmaScript Module
 
 Fluture is written as modular JavaScript (`.mjs`). It can be loaded directly
-by Node 9 and up using `--experimental-modules`, with the [esm loader][esm], or
-with TypeScript (typings included).
+by Node 12 and up using `--experimental-modules`, with the [esm loader][esm].
+Note that the ESM code lives at `fluture/index.mjs`.
```
dicearr:
```suggestion
Note that the ESM code lives at `fluture/index.js`.
```

README.md
```diff
@@ -53,18 +53,37 @@ On older environments you may need to polyfill one or more of the following
...
 
 Besides the module system, no other ES5+ features are used in Fluture's source,
 which means that no transpilation is needed after concatenation.
 
 ```js
 import {readFile} from 'fs'
-import {node, encase, chain, map, fork} from 'fluture'
+import {node, encase, chain, map, fork} from 'fluture/index.mjs'
```
dicearr:
```suggestion
import {node, encase, chain, map, fork} from 'fluture/index.js'
```

PR #390, opened 2019-10-29
Use the built-in TypeScript iterable and generator typings

### LINKED ISSUE #366: Typescript errors for go blocks
I am running a minimal typescript setup using npm. I cannot get the types for go blocks to work. Here is the simple code that I'm running and the compile error it generates:

```typescript
import { resolve, go } from 'fluture'

const func = () => go (function* () {
    return yield resolve (1)
})
```

Error:
```
[eval].ts:2:24 - error TS2345: Argument of type '() => Generator<FutureInstance<unknown, number>, number, FutureInstance<L, any>>' is not assignable to parameter of type 'Generator<FutureInstance<unknown, any>, FutureInstance<unknown, number>>'.
  Type 'Generator<FutureInstance<unknown, number>, number, FutureInstance<L, any>>' is not assignable to type 'Iterator<FutureInstance<unknown, any>, FutureInstance<unknown, number>>'.
    Types of property 'next' are incompatible.
      Type '(...args: [] | [FutureInstance<L, any>]) => IteratorResult<FutureInstance<unknown, number>, number>' is not assignable to type '(value?: FutureInstance<unknown, any>) => Next<FutureInstance<unknown, any>> | Done<FutureInstance<unknown, number>>'.
        Type 'IteratorResult<FutureInstance<unknown, number>, number>' is not assignable to type 'Next<FutureInstance<unknown, any>> | Done<FutureInstance<unknown, number>>'.
          Type 'IteratorYieldResult<FutureInstance<unknown, number>>' is not assignable to type 'Next<FutureInstance<unknown, any>> | Done<FutureInstance<unknown, number>>'.
            Type 'IteratorYieldResult<FutureInstance<unknown, number>>' is not assignable to type 'Done<FutureInstance<unknown, number>>'.
              Property 'done' is optional in type 'IteratorYieldResult<FutureInstance<unknown, number>>' but required in type 'Done<FutureInstance<unknown, number>>'.
```

package.json:
```javascript
{
  "name": "ipfs-playground",
  "version": "1.0.0",
  "main": "index.js",
  "license": "MIT",
  "scripts": {
    "start": "npm run build && node dist/app.js",
    "debug": "npm run build && node --nolazy --inspect-brk=0.0.0.0:9229  dist/app.js",
    "build": "tsc"
  },
  "dependencies": {
    "@types/node": "^12.7.3",
    "fluture": "^12.0.0-beta.3",
    "ts-node": "^8.3.0",
    "typescript": "^3.6.2"
  }
}
```
tsconfig.json:
```javascript
{
    "compilerOptions": {
        "module": "commonjs",
        "target": "ES2015",
        "declaration": true,
        "moduleResolution": "node",
        "sourceMap": true,
        "outDir": "dist",
        "strict": true,
        "baseUrl": "."
    }
}
```

tetsuo:
Fixes #366

This change replaces the Generator/Iterator typings with the built-ins from `es2015.generator` and `es2015.iterable` libs.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,3 +1,6 @@
+/// <reference lib="es2015.generator" />
+/// <reference lib="es2015.iterable" />
+
 declare module 'fluture' {
 
   export interface RecoverFunction {
@@ -18,24 +21,6 @@ declare module 'fluture' {
 
   export interface Nodeback<E, R> {
     (err: E | null, value?: R): void
-  }
-
-  export interface Next<T> {
-    done: boolean
-    value: T
-  }
-
-  export interface Done<T> {
-    done: boolean
-    value: T
-  }
-
-  export interface Iterator<N, D> {
-    next(value?: N): Next<N> | Done<D>
-  }
-
-  export interface Generator<Y, R> {
-    (): Iterator<Y, R>
   }
 
   export interface ConcurrentFutureInstance<L, R> {
@@ -131,7 +116,7 @@ declare module 'fluture' {
   export function forkCatch(recover: RecoverFunction): <L>(reject: RejectFunction<L>) => <R>(resolve: ResolveFunction<R>) => (source: FutureInstance<L, R>) => Cancel
 
   /** Build a coroutine using Futures. See https://github.com/fluture-js/Fluture#go */
-  export function go<L, R>(generator: Generator<FutureInstance<L, any>, R>): FutureInstance<L, R>
+  export function go<L, R>(generator: () => Generator<FutureInstance<L, any>, R>): FutureInstance<L, R>
 
   /** Manage resources before and after the computation that needs them. See https://github.com/fluture-js/Fluture#hook */
   export function hook<L, H>(acquire: FutureInstance<L, H>): (dispose: (handle: H) => FutureInstance<any, any>) => <R>(consume: (handle: H) => FutureInstance<L, R>) => FutureInstance<L, R>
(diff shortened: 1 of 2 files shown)


index.d.ts
```diff
@@ -198,7 +183,7 @@ declare module 'fluture' {
       resolve: ResolveFunction<R>
     ) => Cancel): FutureInstance<L, R>
 
-    'fantasy-land/chainRec'<L, I, R>(iterator: (next: (value: I) => Next<I>, done: (value: R) => Done<R>, value: I) => FutureInstance<L, Next<I> | Done<R>>, initial: I): FutureInstance<L, R>
+    'fantasy-land/chainRec'<L, I, R>(iterator: (next: (value: I) => IteratorYieldResult<I>, done: (value: R) => IteratorReturnResult<R>, value: I) => FutureInstance<L, IteratorYieldResult<I> | IteratorReturnResult<R>>, initial: I): FutureInstance<L, R>
```
Avaq: I think we can replace the union type with [an alias](https://github.com/microsoft/TypeScript/blob/cbbbcfa4c58fb21e5777c8943b18e7ad745dce26/lib/lib.es2015.iterable.d.ts#L41):

```suggestion
    'fantasy-land/chainRec'<L, I, R>(iterator: (next: (value: I) => IteratorYieldResult<I>, done: (value: R) => IteratorReturnResult<R>, value: I) => FutureInstance<L, IteratorResult<L, R>>, initial: I): FutureInstance<L, R>
```

package.json
```diff
@@ -29,7 +29,7 @@
     "test:build": "npm run build && node --experimental-modules --no-warnings -- ./node_modules/.bin/oletus test/build/*.js",
     "coverage:upload": "c8 report --reporter=text-lcov > coverage.lcov && codecov",
     "coverage:report": "c8 report --reporter=html",
-    "test:types": "tsc --lib es6 index.d.ts"
+    "test:types": "tsc index.d.ts"
```
Avaq: It's funny you're making this change now, seeing as it reverts a commit you made:

https://github.com/fluture-js/Fluture/commit/613291d41da3eac8e08c2f759d1680c35f509210

tetsuo: Triple-slash directive already takes care of the `--lib` option

According to https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html

> For example, adding /// <reference lib="es2017.string" /> to one of the files in a compilation is equivalent to compiling with --lib es2017.string.

I checked out that commit which uses TS `v2.8.4`

It doesn't compile without `--lib`

```
> tsc index.d.ts

node_modules/@types/node/util.d.ts(115,88): error TS2304: Cannot find name 'Symbol'.
```

On that branch, some of the dependencies require `estree` which seems to be installing `@types/node` (this should be only installed in development)

`fluture.d.ts` doesn't need `--lib es6` option if I'm correct, if you don't pass that option then TypeScript somehow falls back to `node_modules/@types` (which shouldn't be there in the first place) and `util.d.ts` fails because es6 `Symbol` not defined.

PR #412, opened 2020-01-14
Fix Generator yield type

tetsuo:
Fixes an issue where `go` can't determine the return type of a yield.

```typescript
const g = () => {
  return F.go(function*() {
    const thing: number = yield F.resolve(42)
    return F.resolve(43 + thing)
  })
}
```

([Original bug report on Gitter](https://gitter.im/fluture-js/Fluture?at=5e1d76c443c3b62d79f0d30f))


Avaq:
Thank you Onur! I released this under `12.0.1`. :D

PR #425, opened 2020-05-05
Update sanctuary-type-identifiers to version 3

Avaq:
This update has been a long time coming. I'll explain the reason for waiting so long below, alongside the relevant diff. This update is possibly a breaking change, but I don't see yet how. Maybe @davidchambers can think of a scenario where it would break someone's code. :)

Closes #418 by supersession.

### DIFF
--- a/src/future.js
+++ b/src/future.js
@@ -59,7 +59,13 @@ export function isFuture(x){
   return x instanceof Future || type(x) === $$type;
 }
 
+// Compliance with sanctuary-type-identifiers versions 1 and 2.
+// To prevent sanctuary-type-identifiers version 3 from identifying 'Future'
+// as being of the type denoted by $$type, we ensure that
+// Future.constructor.prototype is equal to Future.
 Future['@@type'] = $$type;
+Future.constructor = {prototype: Future};
+
 Future[FL.of] = resolve;
 Future[FL.chainRec] = chainRec;
 
--- a/src/par.js
+++ b/src/par.js
@@ -23,7 +23,12 @@ export function Par (sequential){
 var $$type = namespace + '/ConcurrentFuture@' + version;
 var zeroInstance = new ConcurrentFuture(never);
 
+// Compliance with sanctuary-type-identifiers versions 1 and 2.
+// To prevent sanctuary-type-identifiers version 3 from identifying
+// 'Par' as being of the type denoted by $$type, we ensure that
+// Par.constructor.prototype is equal to Par.
 Par['@@type'] = $$type;
+Par.constructor = {prototype: Par};
 
 Par[FL.of] = function Par$of(x){
   return new ConcurrentFuture(resolve(x));
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
   ],
   "dependencies": {
     "sanctuary-show": "^2.0.0",
-    "sanctuary-type-identifiers": "^2.0.0"
+    "sanctuary-type-identifiers": "^3.0.0"
   },
   "devDependencies": {
     "c8": "^7.0.0",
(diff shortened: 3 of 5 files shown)


Avaq:
I just realized how this is a breaking change: The removed `Future['@@type']` property was used as a public interface by some future-* libs.

src/future.js
```diff
@@ -59,7 +59,6 @@ export function isFuture(x){
   return x instanceof Future || type(x) === $$type;
 }
 
-Future['@@type'] = $$type;
```
Avaq, line 62: The update to sanctuary-type-identifiers 3 makes it so the removed line above would cause `Future` itself to be identified as a member of `$$type`, as we've also seen in recently https://github.com/fluture-js/fluture-sanctuary-types/pull/67/commits/d8c5527de0d3cdaffc8c47d106d5499290ae581f.

As a result, I am forced to remove this property. I don't want to remove this property, because it means that Fluture no longer supports older code that uses sanctuary-type-identifiers@2 (such as older versions of Fluture, who now lose the ability to understand that Fluture version 12 exposes Future instances.

This is the reason I have been advocating for putting a version number in the property name, eg:

```js
Future['@@type@1'] = 'fluture/Future@3'
Future['@@type@2'] = 'fluture/Future@3'
Future.prototype['@@type@3'] = 'fluture/Future@3'
```

This way, support for multiple versions of a spec can live peacefully alongside each other. I have also advocated for this kind of thing in Fantasy Land, but it's never well received.

I am of the strong opinion that a *spec* must never become outdated. This means that if you make changes to a spec, you must do so in a way that doesn't interfere with the previous version of that spec, which should be treated as a completely different spec.

Avaq:
This PR is no longer breaking because @davidchambers came up with a way to preserve compliance with sanctuary-type-identifiers versions 1 and 2 without interfering with sanctuary-type-identifiers 3: https://github.com/fluture-js/Fluture/commit/d3c2a2578092fa177ca6a92f3ea995a8e565f87f#diff-f9b600f4388a4ec2c76b0b6648a13b44R62-R67

src/future.js
```diff
@@ -59,7 +59,13 @@ export function isFuture(x){
   return x instanceof Future || type(x) === $$type;
 }
 
+// Compliance to sanctuary-type-identifiers version 1 and 2.
```
davidchambers, line 62:
```suggestion
// Compliance with sanctuary-type-identifiers versions 1 and 2.
```

src/par.js
```diff
@@ -23,7 +23,12 @@ export function Par (sequential){
 var $$type = namespace + '/ConcurrentFuture@' + version;
 var zeroInstance = new ConcurrentFuture(never);
 
+// Compliance to sanctuary-type-identifiers version 1 and 2.
```
davidchambers, line 26:
```suggestion
// Compliance with sanctuary-type-identifiers versions 1 and 2.
```

src/future.js
```diff
@@ -59,7 +59,13 @@ export function isFuture(x){
...
+// In order to prevent sanctuary-type-identifiers version 3 from identifying
+// Future as a member of $$type, we ensure that Future.constructor.prototype
+// is equal to Future.
```
davidchambers, line 65: `$$type` is not of type `Type` but of type `String`. :zany_face:

I think `$$type` should be replaced with `Future a b` in the comment above.

The comment could begin with _To prevent_ rather than _In order to prevent_ without any loss of meaning.

Avaq: I'm don't like bringing the type parameters into the picture, because they have nothing to do with sanctuary-type-identifers. I rephrased that part though. What do you think of the current phrasing?

PR #428, opened 2020-05-14
Update type identification logic to understand old Fluture versions

Avaq:
Fluture allows multiple different versions of itself to interact. Historically, whenever a new version of Fluture came out that wouldn't be able to interact with older versions of itself, it would be released as a major new version.

Recently Fluture updated its type identification logic from using sanctuary-type-identifiers 2 to sanctuary-type-identifiers 3. See #399 and #425.

This means that newer versions of Fluture wouldn't understand that they can interact with instances created using older versions of Fluture, and that would be a breaking change.

To prevent it from being a breaking change, this PR adds a fallback for the type identification logic to still be able to identify instances of Future created by older versions of Fluture.

### DIFF
--- a/src/future.js
+++ b/src/future.js
@@ -1,6 +1,5 @@
 /*eslint no-cond-assign:0, no-constant-condition:0 */
-import type from 'sanctuary-type-identifiers';
-
+import {type, legacyType} from './internal/type.js';
 import {FL, $$type} from './internal/const.js';
 import {captureContext, captureApplicationContext, captureStackTrace} from './internal/debug.js';
 import {
@@ -56,7 +55,7 @@ export function Future(computation){
 }
 
 export function isFuture(x){
-  return x instanceof Future || type(x) === $$type;
+  return x instanceof Future || type(x) === $$type || legacyType(x) === $$type;
 }
 
 // Compliance with sanctuary-type-identifiers versions 1 and 2.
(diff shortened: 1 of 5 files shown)


Avaq:
Thank you, @davidchambers. :)

I tested this approach by installing older versions alongside the version from this branch, and running the following code:

```js
> var FL1 = require('fluture')
> var FL2 = require('.')
> FL2.isFuture (FL1.resolve (42))
true
```

Turns out the only versions positively affected by this PR are version 12.0.0 beta 1 through 5. The reason being #399: All Fluture versions since 12.0.0-beta.6 conformed to `sanctuary-type-identifiers@3`.

Future instances from Fluture versions before version 12.0.0-beta.1 are identified to be `fluture/Fluture@4`, and are not compatible with version from after 12.0.0-beta.1, which have identity `fluture/Fluture@5`.

I'm beginning to doubt whether this is worth it. Technically we're still breaking code that is supposed to keep working: Code that relies on mixing instances from Fluture `12.0.0-beta.[1-5]` with instances from Fluture `12.2.x`. However, practically I don't think one would find this anywhere in the wild. :P

davidchambers:
Using a pre-release version of a package is inherently risky. Users know this. I suggest closing this pull request. :)

Avaq:
I think I agree. If I start supporting pre-release versions now, I might be inclined to think twice before making a pre-release next time, which would defeat the purpose of a pre-release in the first place. I'll sleep on it just once, but I'll probably close this branch.

PR #432, opened 2020-06-14
Include copyright licenses in bundled distributions

Avaq:
Fluture distributes several bundles which include its dependencies.

Each of these dependencies are licensed under the MIT license (thank you @davidchambers), so I am allowed to redistribute them.

There is one condition though, and that is that I include the copyright notice in these distributions, which I haven't (sorry @davidchambers).

This PR rectifies that.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,5 +1,5 @@
 The MIT License (MIT)
-Copyright (c) 2018 Aldwin Vlasblom
+Copyright (c) 2020 Aldwin Vlasblom
 
 Permission is hereby granted, free of charge, to any person obtaining a copy of
 this software and associated documentation files (the "Software"), to deal in
--- a/rollup.config.dist.js
+++ b/rollup.config.dist.js
@@ -1,14 +1,31 @@
-/* global process require */
+/* global process require Set */
 
+import {readFileSync} from 'fs';
 import node from '@rollup/plugin-node-resolve';
 import commonjs from '@rollup/plugin-commonjs';
 
 var pkg = require('./package.json');
 
+var dependencies = pkg => {
+  var deps = Object.keys(pkg.dependencies || {}).concat(Object.keys(pkg.peerDependencies || {}));
+  return Array.from(new Set(deps.concat(deps.flatMap(dependency => (
+    dependencies(require(`${dependency}/package.json`))
+  )))));
+};
+
 var banner = `/**
  * Fluture bundled; version ${process.env.VERSION || `${pkg.version} (dirty)`}
  */
 `;
+
+var footer = `/** Fluture license
+
+${readFileSync('./LICENSE')}*/
+
+${dependencies(pkg).map(dependency => `/** ${dependency} license
+
+${readFileSync(`./node_modules/${dependency}/LICENSE`)}*/`
+).join('\n\n')}`;
 
 var typeref = `/// <reference types="https://cdn.jsdelivr.net/gh/fluture-js/Fluture@${
   process.env.VERSION || pkg.version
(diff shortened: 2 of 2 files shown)


rollup.config.dist.js
```diff
@@ -1,15 +1,32 @@
...
 import node from 'rollup-plugin-node-resolve';
 import commonjs from 'rollup-plugin-commonjs';
 
 var pkg = require('./package.json');
 
+var dependencies = pkg => {
+  var deps = Object.keys(pkg.dependencies || {}).concat(Object.keys(pkg.peerDependencies || {}));
+  return Array.from(new Set(deps.concat(deps.flatMap(dependency => (
+    dependencies(require(`${dependency}/package.json`))
+  )))));
```
davidchambers, line 13: Should this array be sorted in some way?

rollup.config.dist.js
```diff
@@ -1,15 +1,32 @@
...
  */
 `;
 
+var footer = `/** Fluture license
+
+${readFileSync('./LICENSE')}*/
+
+${dependencies(pkg).map(dependency => `/** ${dependency} license
+
+${readFileSync(`./node_modules/${dependency}/LICENSE`)}*/`
```
davidchambers, line 27: Is it safe to assume that the file is always named __LICENSE__?

Avaq: No. There are several unsafe assumptions in the code above, like the assumption that the dependency is available directly under `node_modules`. However, for my current list of dependencies it works fine. I figured I'd tackle those problems in the unlikely event that they'd come up.

Maybe I should just remove the code completely and manually specify a list of license files to include.

rollup.config.dist.js
Avaq: I don't see why. But I'm considering listing them out myself instead of generating them like this, in which case I can choose any order I'd like.

Avaq:
@davidchambers How do you view this change at the conceptual level?

Avaq:
I'm asking because the license conditions for JavaScript libraries are very commonly ignored without consequences. Just think of all the WebPack -created web apps that bundle thousands of packages with various open source licenses and distribute those bundles to thousands of computers without any kind of attribution. Such is the nature of the web.

The lines also get fairly blurry fairly quickly. Imagine I don't bundle my dependencies, but host them as-is on my web server. This might become more common in the future as HTTP2 and ESM gain traction. The license file is neatly hosted as well, alongside the JavaScript. But a user agent only ever fetches the JavaScript file, because `index.js` contains `import S from './web_modules/sanctuary.js'`. Now did I do enough as a developer to adhere to the license terms?

rollup.config.dist.js
davidchambers: Listing the dependencies would give you greater flexibility, but would also create a maintenance problem. I like your current approach provided that there is no risk of bungling a release if and when your assumption is violated.

rollup.config.dist.js
davidchambers: I don't like to rely on the internal ordering of unordered data structures. Even if the ordering is deterministic, it may appear arbitrary. Additionally, the relative ordering of elements may change when an element is added to the set, which could result in noisy diffs.

I don't know whether these concerns apply in this particular situation.

davidchambers:
> How do you view this change at the conceptual level?

Very positively. You are doing what you can to respect the licences, which is laudable. :)

rollup.config.dist.js
Avaq: I don't think it matters for the licenses to be listed in arbitrary order.

LICENSE
```diff
@@ -1,5 +1,5 @@
 The MIT License (MIT)
-Copyright (c) 2018 Aldwin Vlasblom
+Copyright (c) 2020 Aldwin Vlasblom
```
davidchambers, line 2: I see that this project does not yet use sanctuary-scripts. ;)

Issue #331, opened 2019-04-21
Improve the promise function


Avaq:
The current function signature is not true to reality:

```hs
promise :: Future a b -> Promise b a
```

The rejection branch of a Promise is not considered "typed" in the same way that `catch (e)` statements are untyped. Anything can be thrown. The `Promise b a` in the signature there is made up. Typically Promises are considered unary types. Practically users have come to expect the rejection branch to contain an instance of `Error`, similarly to `catch (e)`.

To uphold this signature, however, Fluture is forced to `throw` its exception whenever `promise` was called on a crashed Future. A better flow would be to pass the exception into the Promise' rejection branch, in the spirit of Promises.

To accommodate for this change, I propose changing the signature of `promise`:

```hs
promise :: Future Error a -> Promise a
```

We can see what's happening here:

1. We're forcing users to call `Promise` on Futures that have *Error* in their rejection branch *only*.
2. Now both the exception branch *and* the rejection branch can be merged into the rejection branch of the Promise, in the spirit of Promises.
3. Our Promise type is no longer made up.

---

Since merging user-land errors with runtime exceptions is generally a bad idea, I will urge users to use `fold` before `promise` to keep rejections separate from crashes. If users have a non-`Error` type in their rejection branch, they will be forced to do this anyway.

---

I think this change caters to two audiences at the same time:

1. Those who were already using the rejection branch for Error instances and don't mind merging them with runtime exceptions when casting to a Promise.
2. Those who keep their rejections separate from their crashes, but were not able to use `promise` because their crashes would've been thrown.

edgarjrg:
In `fluture@12.0.0-beta.3` the type of `promise` is 
```ts
export function promise<R>(source: FutureInstance<Error, R>): Promise<R>
```

That makes very cumbersome the transformation from Fluture to Promise, I get the reasoning, but many libraries/users do not put an `Error` in `Promise` rejection branch, and typescript lib.es5 types `catch` as 

```ts
catch<TResult = never>(onrejected?: ((reason: any) => TResult | PromiseLike<TResult>) | undefined | null): Promise<T | TResult>;
```

So the reason is `any` in either way.

Maybe we should be more flexible and type it 

```ts
  export function promise<R>(source: FutureInstance<any, R>): Promise<R>
```

Avaq:
I understand where you're coming from. If all Fluture would do is move the rejection branch from a Future into the Promise rejection branch, then I think it's a good suggestion.

[...]

PR #434, opened 2020-07-07
`promise()` declaration should accept any rejection type

wms:
Constraining the left-hand-side of the input Future to the Error type makes no sense, since Futures (and indeed Promises, although that's not applicable here) can reject with any type, not necessarily an Error.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -158,7 +158,7 @@ export function resolve<R>(value: R): FutureInstance<never, R>
 export function parallel(concurrency: number): <L, R>(futures: Array<FutureInstance<L, R>>) => FutureInstance<L, Array<R>>
 
 /** Convert a Future to a Promise. See https://github.com/fluture-js/Fluture#promise */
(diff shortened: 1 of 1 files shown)


Avaq:
Hi @wms, and thank you for your contribution.

However, a lot of thought went into the design of the `promise` function and its current type. Most of it has been documented in #331. If you have the time, please have a look at this thread, as it explains in detail why the left hand side of the Future is constrained.

For a summary, you can refer to the Fluture 12.x upgrade guide, which contains [a note about the new design of the `promise` function](https://gist.github.com/Avaq/ee2c6c819db4c37258e9a226e6380a38#promise-function-reworked).

Let me know what you think. :)

wms:
@Avaq Thank you for the response; I'm surprised I'd not seen that prior as I spent a bit of time grepping around to figure out if anybody had brought this up already. I'll digest it and get back with my thoughts.

wms:
I dig it; I'll close this out.

PR #439, opened 2020-08-11
Add conditional exports data to the package

Avaq:
This allows Node 14 users to import the modular code straight from "fluture", as opposed to having to deep import from "fluture/index.js".

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,8 +68,9 @@ single files that come with all of Fluture's dependencies pre-bundled.
 
 Fluture is written as modular JavaScript.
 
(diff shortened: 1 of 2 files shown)


Avaq:
This might need to be considered a breaking change. Prior to it, if a Node 14 user would do:

```js
import Future from 'fluture'

Future.resolve (42)
```

That would work. Because `'fluture'` would resolve to the CommonJS module, which contains all functions.

Now, this will throw `TypeError: Future.resolve is not a function`, because instead of loading the commonjs export, we now load the default modular export from `index.js`, which is just the Future constructor without any additional properties attached.

dicearr (approved):

PR #442, opened 2020-08-20
Add missing exports

dicearr:
While updating Momi I realized that it is not possible to deep import any file which is not exported in the `exports` object.

```text
Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './test/assertions.js' is not defined by "exports"
in /home/diego/dev/fluture/momi/node_modules/fluture/package.json imported
from /home/diego/dev/fluture/momi/test/index.js
```

I have only added  `test/assertions` and `test/arbitraries` but I don't know what to do with the types.

I wouldn't export `src/*` although I am not aware if we have any important project deep importing from it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,6 +12,7 @@
       "import": "./index.js",
       "default": "./index.cjs"
     },
+    "./index.js": "./index.js",
     "./test/arbitraries.js": "./test/arbitraries.js",
     "./test/assertions.js": "./test/assertions.js"
   },
--- a/package.json
+++ b/package.json
@@ -7,8 +7,12 @@
   "types": "index.d.ts",
   "module": "index.js",
   "exports": {
-    "require": "./index.cjs",
-    "import": "./index.js"
+    ".": {
+      "require": "./index.cjs",
+      "import": "./index.js"
+    },
+    "./test/arbitraries.js": "./test/arbitraries.js",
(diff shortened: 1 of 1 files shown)


Avaq:
> While updating Momi I realized that it is not possible to deep import any file which is not exported in the exports object.

In that case I'd also like to add `./index.js` to preserve backwards compatibility.

Avaq:
> I don't know what to do with the types.

I don't know what you mean.

dicearr:
Well, does types need to be imported?

dicearr:
> > While updating Momi I realized that it is not possible to deep import any file which is not exported in the exports object.
> 
> In that case I'd also like to add `./index.js` to preserve backwards compatibility.

I will add it. So, it wouldn't have been a breaking change.

Avaq:
> So, it wouldn't have been a breaking change.

It still is for users who were doing `import 'fluture'`. Just no longer for users who were doing `import 'fluture/index.js'`.

Avaq:
> Well, does types need to be imported?

You mean the TypeScript types? I don't think TypeScript cares about the "exports" at all. It does its own thing. TypeScript has always imported the modular code, even when "main" points somewhere else.

Avaq (approved):

PR #457, opened 2021-01-07
Fix TypeScript type inference for functions passed to map

### LINKED ISSUE #455: Unable to infer types in function passed to `map`
Hello,

I am still using fluture 11.x in my biggest projects, and have been starting to dip my toes into v13. I may be missing something, but there seems to be an annoying bug where the function I pass to `map` cannot infer types.

To reproduce:

tsconfig.json
```json
{
    "include": [
        "src/**/*",
        "test/**/*"
    ],
    "exclude": [
        "node_modules",
        "dist"
    ],
    "compilerOptions": {
        "target": "es5",
        "module": "commonjs",
        "lib": ["es6"],
        "declaration": true,
        "declarationMap": true,
        "sourceMap": true,
        "outDir": "./dist",
        "downlevelIteration": true,
        "strict": true,
        "baseUrl": "./",
        "paths": {
            "*": ["src/@types/*"]
        },
        "esModuleInterop": true
    }
}

```

package.json
```package.json
{
  "name": "repro-fluture",
  "version": "0.0.1",
  "devDependencies": {
    "@types/node": "^14.14.17",
    "ts-node": "^9.1.1",
    "typescript": "^4.1.3"
  },
  "dependencies": {
    "fluture": "^13.0.1"
  }
}

```

src/index.ts
```typescript
import * as F from 'fluture'

F.resolve(5)
    .pipe(F.map(value => value.toString()))
```

This code gives the error:

```
src/repro-fluture-13-map-type-bug.ts:4:26 - error TS2571: Object is of type 'unknown'.

4     .pipe(F.map(value => value.toString())
                           ~~~~~
```

Here the type of `value` is `unknown`, so the `toString()` cannot be type-checked.

I find it strange that nobody has reported this, but I am able to reproduce in several of my repos, with various tsconfig settings.

Avaq:
This pull request tries to fix #455 

I've started with a [tsd](https://github.com/SamVerschueren/tsd) setup to defend against regressions to before #401 and #403. Run `npm run test:types` to check that your typings have not introduced a regression.

For context, here's the current TypeScript typings for `map`:

https://github.com/fluture-js/Fluture/blob/2f12660b744b28b3ed486f3dfef6bdf233dbd814/index.d.ts#L137-L143

Also relevant, the type of `pipe`:

https://github.com/fluture-js/Fluture/blob/2f12660b744b28b3ed486f3dfef6bdf233dbd814/index.d.ts#L36-L37

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -28,7 +28,7 @@ export interface Functor<A> {
   'fantasy-land/map'<B extends this[typeof $T]>(mapper: (value: A) => B): Functor<B>
 }
 
-type Unfunctor<F extends Functor<unknown>, B> = ReturnType<(F & { [$T]: B })['fantasy-land/map']>
+type Mapped<F extends Functor<unknown>, B> = ReturnType<(F & { [$T]: B })['fantasy-land/map']>
 
 export interface ConcurrentFutureInstance<L, R> extends Functor<R> {
   sequential: FutureInstance<L, R>
@@ -145,8 +145,8 @@ export function lastly<L>(cleanup: FutureInstance<L, any>): <R>(action: FutureIn
 
 /** Map over the resolution value of the given Future or ConcurrentFuture. See https://github.com/fluture-js/Fluture#map */
 export const map: {
-  <B, F extends Functor<unknown>>(mapper: Functor<unknown> extends F ? never : (a: F extends Functor<infer A> ? A : never) => B): (source: F) => Unfunctor<F, B>
-  <A, B>(mapper: (a: A) => B): <F extends Functor<A>>(source: F) => Unfunctor<F, B>
+  <B, F extends Functor<unknown>>(mapper: Functor<unknown> extends F ? never : (a: F extends Functor<infer A> ? A : never) => B): (source: F) => Mapped<F, B>
+  <A, B>(mapper: (a: A) => B): <F extends Functor<A>>(source: F) => Mapped<F, B>
 }
 
 /** Map over the rejection reason of the given Future. See https://github.com/fluture-js/Fluture#maprej */
--- a/index.d.ts
+++ b/index.d.ts
@@ -145,8 +145,8 @@ export function lastly<L>(cleanup: FutureInstance<L, any>): <R>(action: FutureIn
 
 /** Map over the resolution value of the given Future or ConcurrentFuture. See https://github.com/fluture-js/Fluture#map */
 export const map: {
-  <B, F extends Functor<unknown>>(f: Functor<unknown> extends F ? never : (a: F extends Functor<infer A> ? A : never) => B): (source: F) => Unfunctor<F, B>
-  <A, B>(f: (a: A) => B): <F extends Functor<A>>(f: F) => Unfunctor<F, B>
+  <B, F extends Functor<unknown>>(mapper: Functor<unknown> extends F ? never : (a: F extends Functor<infer A> ? A : never) => B): (source: F) => Unfunctor<F, B>
+  <A, B>(mapper: (a: A) => B): <F extends Functor<A>>(source: F) => Unfunctor<F, B>
 }
 
 /** Map over the rejection reason of the given Future. See https://github.com/fluture-js/Fluture#maprej */
--- a/index.d.ts
+++ b/index.d.ts
@@ -21,17 +21,19 @@ export interface Nodeback<E, R> {
   (err: E | null, value?: R): void
 }
 
+declare const $T: unique symbol
+
 export interface Functor<A> {
-  input: unknown
-  'fantasy-land/map'<B extends this['input']>(mapper: (value: A) => B): Functor<B>
+  [$T]: unknown
+  'fantasy-land/map'<B extends this[typeof $T]>(mapper: (value: A) => B): Functor<B>
 }
 
-type Unfunctor<F extends Functor<unknown>, B> = ReturnType<(F & { input: B })['fantasy-land/map']>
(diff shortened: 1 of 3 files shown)


Avaq:
I've pushed an additional set of test assertions to check that the requirements stipulated in #455 are met. As of yet, these assertions fail and I have been unable to find a type definition that causes the assertions to pass.

Avaq:
Of course, if I replace the entire TS definition for `map` with the following, then the new assertions pass:

```ts
export function map<RA, RB>(mapper: (value: RA) => RB):
  <L>(source: FutureInstance<L, RA>) => FutureInstance<L, RB>
```

The definition above is functionally equivalent to the definition that was used in Fluture v11. There is no support for the ConcurrentFuture type, and so while all the other assertions now pass, the ones for the ConcurrentFuture type fail (#401).

index.d.ts
```diff
@@ -21,20 +21,27 @@ export interface Nodeback<E, R> {
   (err: E | null, value?: R): void
 }
 
-export interface ConcurrentFutureInstance<L, R> {
+export interface Functor<A> {
+  input: unknown
```
tjjfvi, line 25: I recommend changing `input` to a unique symbol, to prevent it from showing up in autocomplete:
```ts
declare const __functorInput: unique symbol;
export interface Functor<A> {
  [__functorInput]: unknown
  'fantasy-land/map'<B extends this[__functorInput]>(mapper: (value: A) => B): Functor<B>
}
```
You should also export the symbol if you want to support external `Functor` subtypes with `map`.

Avaq: Hey, I was thinking exactly this while going to sleep yesterday! I will make this change. Since it's a symbol I don't need it to have a obscure name, right?

Avaq: Another thought about the name: It's a placeholder for the yet-to-be-known type of `B` in `Functor<B>`, which is the *output* of the map operation, so calling it *input* may be misleading. But I'm guessing that when viewed from a different perspective, the name `input` does make sense?

Avaq: Actually this proposal doesn't seem to do the correct thing. Firstly, I need to use `[typeof __functorInput]`, because `__functorInput` is seen as a value, but when I do that, I guess the type becomes `unique symbol`, which is not specific enough, because we need the exact constant symbol that we've defined, as happens when using a constant string.

Avaq: It looks like there's no thing like a "symbol literal": https://www.staging-typescript.org/docs/handbook/literal-types.html

Avaq: Actually, my mistake. It does work with `typeof __functorInput`. I just forgot to change it in one occurrence.

Avaq: > to prevent it from showing up in autocomplete

It does have that effect. :)

> You should also export the symbol if you want to support external Functor subtypes with map.

Once I do, it starts showing up in autocomplete again, leading it to auto-import the symbol. :P

index.d.ts
```diff
@@ -21,20 +21,29 @@ export interface Nodeback<E, R> {
   (err: E | null, value?: R): void
 }
 
-export interface ConcurrentFutureInstance<L, R> {
+declare const $functor: unique symbol
```
Avaq, line 24: I named this after the interface it's used in. But in theory I could use this symbol in other similar interfaces, right? Maybe I should give it a more general name.

Avaq: Maybe just `$T`? It just functions as an injection point for type information, right?

index.d.ts
```diff
@@ -21,20 +21,29 @@ export interface Nodeback<E, R> {
...
 
-export interface ConcurrentFutureInstance<L, R> {
+declare const $T: unique symbol
+
+export interface Functor<A> {
+  [$T]: unknown
+  'fantasy-land/map'<B extends this[typeof $T]>(mapper: (value: A) => B): Functor<B>
+}
+
+type Unfunctor<F extends Functor<unknown>, B> = ReturnType<(F & { [$T]: B })['fantasy-land/map']>
```
tjjfvi, line 31: I don't think this is really `Unfunctor`; I think it's more like `Refunctor`, as it takes in `<SomeFunctor<A>, B>` and returns `SomeFunctor<B>`.

Also, it might be worthwhile to add
```ts
type Unfunctor<F extends Functor<unknown>> = F extends Functor<infer T> ? T : never;
```
and to use that in the first line of `map`; atm it's a pretty long, hard-to-read line.

index.d.ts
tjjfvi: `input` makes sense from the HKT standpoint, so that's why I chose it initially. `$T` works well though IMO.

index.d.ts
Avaq: Good ideas. I wasn't too happy about the Unfunctor name either, but I find that thing hard to name.

Avaq: Maybe `Mapped`?

Avaq: > it might be worthwhile to add [an alias for inferring the inner value of a Functor]

Actually, it seems it's only ever going to be used in one place.

Avaq:
Should I release this with a major version bump? I would prefer not to, but I find that when changing TS types, it usually probably breaks something for someone.

Avaq:
We've kind of "narrowed" the type of `map` here from `map(f: (unknown) => B)` to `map(f: (A) => B)` where `A` can be inferred from the second argument. Narrowing a type is usually a breaking change. In this case, we're narrowing the type to exactly what the user would want, though. So chances are users will have no problems. Unless a user gave an incorrectly typed function to `map`, then this change will suddenly break their code. Although in the latter case one might argue that their code was already broken, they just didn't know it.

